PR #11, opened 2021-04-02
typo in README 

hicksy:
The following refers to an incorrect (old maybe?) way to create a model 

```javascript
const Card = new Model(table, {
    name: 'Card',
    fields: { /* Model schema field definitions */ }
})
```

This should be

```javascript
const Card = new Model(table, 'Card', {
    fields: { /* Model schema field definitions */ }
})
```


mobsense:
Thank you.

PR #28, opened 2021-04-29
Initialize typeField value for find method.

mkovel:
It allows do: Model.find({ },{ }) without pasing Model.find({ _type: 'Role',  },{ });

````javascript
const RoleModel = new Model(table, 'Role', {
  fields: {
    pk: { value: '${_type}' },
    sk: { value: '${_type}:${id}' },
    id: { type: String, ulid: true, required: true },
    name: { type: String, required: true },
    permissions: { type: Array, required: true },
  },
});

const roles = await RoleModel.find({ },{ limit: 3, parse: true });
`

### DIFF
--- a/src/Model.js
+++ b/src/Model.js
@@ -543,6 +543,8 @@ export class Model {
     }
 
     /* private */ async queryItems(properties = {}, params = {}) {
+        properties = Object.assign({}, properties)
+        properties[this.typeField] = this.name        
         let expression = new Expression(this, 'find', properties, params)
         return await this.run('find', expression)
     }

### CODE REVIEW

**Comment by mobsense** [comment]
One comment:

This change will generally add the "type" to the FilterExpression list for a small (perhaps not measurable) performance hit inside DynamoDB itself. When you have the type as part of the PK / SK, you don't need the filter expression.

To prevent this, you can add the "type" to the schema and set filter to false.

ie.

```
  fields: {
    pk: { value: '${_type}' },
    sk: { value: '${_type}:${id}' },
    type: { type: String, filter: false },
  }
```

This can be done if you never want to filter on type in a way that cannot be done via the PK/SK.

A few other suggestions:

1. It is often best to centralize your schema rather than providing it via fields. i.e. provide to the Table constructor
2. Best to add a "type" to the PK/SK. It does default to "type: String", but probably best to be explicit. We've updated our examples in this regard.
3. parse: true should be the default

PR #47, opened 2021-06-13
move typeField from low level queryItems to find

gavrilyak:
I tried this example from readme
```
//  Fetch an item collection for Acme
let items = await table.queryItems({pk: 'account:AcmeCorp'}, {parse: true})

//  Group items into arrays by model type
items = db.groupByType(items)
```
and it doesn't work, because queryItems applied filter by _type = "Generic"

It works with scanItems though.

So my proposed patch fixes it for queryItems too.

Probably the same needs to be done for putItem and updateItem, but it's harder for me right now.

Thanks for great project!

### DIFF
--- a/src/Model.js
+++ b/src/Model.js
@@ -463,6 +463,8 @@ export class Model {
     async find(properties = {}, params = {}) {
         this.checkArgs(properties, params)
         params = Object.assign({parse: true, high: true}, params)
+        properties = Object.assign({}, properties)
+        properties[this.typeField] = this.name
         return await this.queryItems(properties, params)
     }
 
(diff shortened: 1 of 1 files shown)


mobsense:
Thanks for finding and reporting.

I think there may be problems with your proposed solution.

queryItems also needs the type when called directly from a model. So moving the type from `queryItems` to `find` will break Model.queryItems.

We've got a solution by setting a Model.generic option and testing that in queryItems, updateItems and putItems. If set, then type type is not defined.

Also found an issue with groupByType where the iteration of items wasn't correct. 

Will commit these changes after test later today.

PR #82, opened 2021-07-29
fix: wrong path to dist file

radekl:
Not sure how it worked a month ago, but right now when trying to run app through serverless it throws me an error that `main` is not found. Shouldn't this path be similar as in `mjs` case?

### DIFF
--- a/Dynamo/package.json
+++ b/Dynamo/package.json
@@ -1,7 +1,7 @@
 {
     "name": "Dynamo",
     "description": "AWS SDK V3 access for OneTable",
-    "main": "dist/cjs/Dynamo.js",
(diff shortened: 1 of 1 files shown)


radekl:
It turned out to be a problem with node `v12.16.1` - in that version it was trying to load `cjs`. `v14.16.0` it works properly.

mobsense:
Thank you for digging into this. Are you saying this patch is required or is the pull request not required if using Node 14.  Or are you saying we need the patch?

radekl:
Sorry for late response:
- It is not required for Node v14
- Without this change it doesn't work on Node v12

I have checked locally and this change did not affect v14.

PR #90, opened 2021-08-06
chore: moved missing runtime deps

radekl:
Moved devDependencies that are actually used during runtime. Lack of it causes apps that are using this library to fail due to missing dependencies. Without this fix it works on developer machine.

Haven't tested it that way (it failed when I deployed app to AWS Lambda) is to use `npm install --production` on app that uses `dynamodb-onetable` locally. I have verified (manually and on limited scope) that this change helps.


mobsense:
Thanks, looks good.

PR #97, opened 2021-08-16
Enhancement: Allow local forks to use docker to run dynamodb-local

simonireilly:
Morning @mobsense, and wow, thanks for this awesome library, I really enjoyed reading the docs and seeing the approach.

Your check-list for dynamoDB has gone straight onto my list of tools in notion :pray: 

## Description

- Adds docker support for running tests; this boots docker instead of a local dynamo-db for people without Java
- Adds a `.prettierrc` which enforces the documented coding standard in `CONTRIBUTING.md`
    >Make sure your code style matches existing code. Soft tabs=4, no trailing spaces etc.
- Adds directions for install, build, test, and lint for contributors.

Please, let me know any feedback, and thanks for this tool :+1:

## Demonstration

1. Setup a watcher to see what docker spawns
2. Run tests
3. Show docker spinning down afterward

docker-running-tests

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,6 @@
+{
+  "trailingComma": "es5",
+  "tabWidth": 4,
+  "semi": false,
+  "singleQuote": true
+}
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,6 +1,0 @@
-{
-  "trailingComma": "es5",
-  "tabWidth": 4,
-  "semi": false,
-  "singleQuote": true
-}
--- a/package-lock.json
+++ b/package-lock.json
@@ -2878,6 +2878,12 @@
                 "node": ">= 0.8"
             }
         },
+        "node_modules/commander": {
+            "version": "3.0.2",
+            "resolved": "https://registry.npmjs.org/commander/-/commander-3.0.2.tgz",
+            "integrity": "sha512-Gar0ASD4BDyKC4hl4DwHqDrmvjoxWKZigVnAbn5H1owvm4CxCPdb0HQDehwNYMJpla5+M2tPmPARzhtYuwpHow==",
+            "dev": true
+        },
         "node_modules/component-emitter": {
             "version": "1.3.0",
             "resolved": "https://registry.npmjs.org/component-emitter/-/component-emitter-1.3.0.tgz",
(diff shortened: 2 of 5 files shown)


.prettierrc
```diff
@@ -0,0 +1,6 @@
+{
```
simonireilly, line 1: Attempted to use the styles already prominent in the code base, to make contributing follow style guide by default.

package-lock.json
```diff
@@ -1,7722 +1,8 @@
 {
     "name": "dynamodb-onetable",
     "version": "1.7.2",
-    "lockfileVersion": 2,
```
simonireilly, line 4: Let me know what happened here, I just ran`npm install wait-port --save-dev` and a lot seemed to change

package.json
```diff
@@ -62,7 +62,8 @@
         "eslint": "^7.25.0",
         "jest": "^26.6.3",
         "ts-jest": "^26.4.4",
-        "typescript": "^4.1.3"
+        "typescript": "^4.1.3",
+        "wait-port": "^0.2.9"
```
simonireilly, line 66: New dependency, it seems well maintained, using it to detect when docker port opens after spinning up the spawned process.

test/utils/setup.ts
```diff
@@ -1,14 +1,50 @@
...
+    } else {
+        console.info('Using local Java to run dynamoDB')
+        dynamodb = DynamoDbLocal.spawn({ port: PORT })
+    }
+
+    process.env.DYNAMODB_PID = String(dynamodb.pid)
+    process.env.DYNAMODB_PORT = String(PORT)
+
+    // When jest throws anything unhandled, ensure we kill the spawned process
+    process.on('unhandledRejection', (error) => {
```
simonireilly, line 42: This should ensure we don't leave the spawned process running in the event of anything nasty happening.

mobsense:
Thanks for the pull request. Let me look it over and get back here.

test/utils/setup.ts
mobsense: Would it be better to pass in process.env.DOCKER the command line so the port would not be hard coded?

.prettierrc
mobsense: Sounds good. Does the current code follow those specs in all cases?  I know we do something a little different with switch/case where we out-dent the case statements.

package-lock.json
mobsense: Lockfiles can change wildly with updates.

mobsense:
Overall looks pretty good. Can you think about passing in the DOCKER command line via the env. 

AND: sorry to be a pain, could you split the prettyrc into a separate pull request. We try to have each pull request focus on a single feature.

.prettierrc
simonireilly: Yep, split out :+1:

test/utils/setup.ts
```diff
@@ -1,14 +1,50 @@
...
 
 module.exports = async () => {
+    let dynamodb
 
-    let dynamodb = DynamoDbLocal.spawn({port: PORT})
-    process.env.DYNAMODB_PID = dynamodb.pid
-    process.env.DYNAMODB_PORT = PORT.toString()
-    // console.log('Spawn DynamoDB', dynamodb.pid)
+    if (String(process.env.DOCKER) === 'true') {
+        const args = [`run`, `-p`, `${PORT}:8000`, `amazon/dynamodb-local`]
```
simonireilly, line 14: re: https://github.com/sensedeep/dynamodb-onetable/pull/97/files#r689922361

Internally on the docker network, 8000/tcp will be running for the container, but here it can be mapped to whatever port is set as `PORT`, on the host machine.

This means that it works the same way as:

```ts
dynamodb = DynamoDbLocal.spawn({ port: PORT })
```

See my suggestion above :+1:

test/utils/setup.ts
```diff
@@ -1,14 +1,50 @@
 /*
     Setup -- start dynamodb instance
  */
+import { spawn } from 'child_process'
+const waitPort = require('wait-port')
 import DynamoDbLocal from 'dynamo-db-local'
 
 const PORT = 4567
```
simonireilly, line 8:
```suggestion
const PORT = parseInt(process.env.DYNAMODB_PORT || 4567)
```

package-lock.json
simonireilly: I think I might be using npm 6, whilst this `package-lockfile.json` was created with npm 7; I had npm 6 locally.

The key for me was that it actually was decreasing the lockfile version, which really did not look right to me.

I did:

```
npm install -g npm@latest

npm install
```

the diff should be far smaller

test/utils/setup.ts
simonireilly: Answered in https://github.com/sensedeep/dynamodb-onetable/pull/97/files#r690077693; let me  know, I can always make the change, I am  just trying to understand it :+1:

package-lock.json
mobsense: Yes, npm 6 actually fails on the Dynamo.js file. We need npm 7.  Can't wait till the next node goes LTS with NPM 7.

test/utils/setup.ts
mobsense: Could we get to just one env var?  i.e

DYNAMO_DOCKER_PORT?

mobsense: Could we get rid of the DOCKER=true and trigger off the DYNAMODB_PORT?

package-lock.json
simonireilly: A candidate for a separate merge could be setting engines in the package.json to specify minimum node versions and npm versions:

```json
"engines": {
    "node": ">=12.0.0",
    "npm": ">=7.0.0"
  },
```
This would give a friendly error I think

mobsense: Yes please !!

test/utils/setup.ts
simonireilly: I think I understand, does this make sense:

docker-running-tests-v2

This would mean that we can have long-running local dynamodb, without worrying about port collision when running tests, is that the idea?

mobsense: Yes. You may need to do the PORT references on line 8 and 45?

The only other possibility to consider would be to call it DYNAMODB_DOCKER and you could set that to a port or true. If true, then use an internal pre-defined port. Otherwise, it gives the port number.

That way we can document for most people to use DYNAMODB_DOCKER=true but others can do DYNAMODB_DOCKER=17070 or similar.

What I wanted was to default to one env var (PID excepted) and be as simple as possible.

Thanks for your work on this. Great to have your help.

simonireilly: Ok, I think I have some comments on this how we might go forward to make this straightforward, so, please let me know what you think.

1. User can specify PORT that local dynamodb runs on
2. User can specify DOCKER as a truthy value 'yes', 'true', '1'

## Examples

```bash
npm test 
> port = 4567, does not use docker
PORT=1234 npm test
> port = 1234, does not use docker
DOCKER=true npm test
> port = 4567, uses docker
DOCKER=true PORT=1234 npm test
> port = 1234, uses docker
```

In this way, we separate concerns:

- Do you want docker (yes/no)
- Do you want local dynamoDB on a specific port (provide a port number)

## Reason

The user has the same experience, regardless of docker or not, they just get a port that points at local docker.

Let me know what you think :thinking:

mobsense: Sounds good. Sorry to give you a bad steer on this.

mobsense: Let me know when you think this is complete.

mobsense:
You may need to update to fix the package-lock issue before I can merge.

simonireilly:
Yeah, I will pull and rebase onto master, then push :+1:

test/utils/setup.ts
simonireilly: Happy with this one, going to resolve :+1:

mobsense:
Thank you!!  I need to build my docker skills and explore deploying using docker vs zip for serverless.  So much to do ... so little time.  Thanks.

PR #99, opened 2021-08-17
Maintenance: Specify node, and npm versions

simonireilly:
Based on: https://github.com/sensedeep/dynamodb-onetable/pull/97#discussion_r690089307

## Changes

- Ensure that npm is version 7
- Ensure that node is at least version 12.x

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,5 +67,9 @@
     "files": [
         "dist/",
         "Dynamo"
-    ]
+    ],
+    "engines": {
+        "node": ">=12.0.0",
+        "npm": ">=7.0.0"
+    }
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,6 +21,10 @@
                 "jest": "^26.6.3",
                 "ts-jest": "^26.4.4",
                 "typescript": "^4.1.3"
+            },
+            "engines": {
+                "node": ">=12.0.0",
+                "npm": ">=7.0.0"
             }
         },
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -67,5 +67,9 @@
     "files": [
         "dist/",
         "Dynamo"
-    ]
+    ],
+    "engines": {
+        "node": ">=12.0.0",
```
simonireilly, line 72: @mobsense; I expect lots of people are using this library with node 12, but I believe from the docs certain features only work with node 14. 

Should I make this greater than or equal to node 14?

mobsense: I can't remember being dependent on node 12. We do require NPM 7 if you want to use AWS SDK V3.

Because we transpile, we should not be requiring node 14.

simonireilly: Ok super, my guess would be that it makes sense to add node 12.x to the matrix of tests?

Here is a link to my fork where I have made that change: https://github.com/simonireilly/dynamodb-onetable/pull/1

That one could go in separately as well? What do you think?

mobsense: Yes, makes sense.

mobsense:
Thank you for thinking more broadly about our testing. Always good.

simonireilly:
> Thank you for thinking more broadly about our testing. Always good.

No problem, thanks for giving this your attention today, I appreciate your time :+1:

PR #104, opened 2021-08-21
FIX: Handle interpolated values in createUnique

sdemjanenko:
When using an interpolated value as a unique attribute, this method was
saving `undefined` on the `pk`. This commit updates the logic to run the
interpolation before extracting the unique field value from properties.

This commit also adds a test for this case.

NOTE: It isn't clear that `new Expression` is required. Comments above
`prepareProperties` suggest that it is, but manual testing suggests
otherwise.

### DIFF
--- a/src/Model.js
+++ b/src/Model.js
@@ -527,8 +527,13 @@ export class Model {
         let {hash, sort} = this.indexes.primary
         let fields = this.block.fields
         fields = Object.values(fields).filter(f => f.unique && f.attribute != hash && f.attribute != sort)
+
+        let preparedProperties = this.prepareProperties('put', properties, params);
+        // Is expression necessary here? In my testing it isn't but prepareProperties suggests its necessary
+        let uniqueExpression = new Expression(this, 'put', preparedProperties, params);
         for (let field of fields) {
-            await this.table.uniqueModel.create({pk: `${this.name}:${field.attribute}:${properties[field.name]}`}, {
+            // TODO: Switch to using the table delimiter
+            await this.table.uniqueModel.create({pk: `${this.name}:${field.attribute}:${uniqueExpression.properties[field.name]}`}, {
                 transaction,
                 exists: false,
                 return: 'NONE',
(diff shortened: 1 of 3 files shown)


mobsense:
Thanks. Let me check this out.

It is definitely desirable to have value templates for unique properties.  AND ...

the hard-coded ":" is a bug. 

Looking at the format of the unique:

```
_Unique:MODEL:ATTRIBUTE:VALUE
```

it is possible, but extremely unlikely that someone could encode data values with ":" when they are using "#" as a delimiter and this could clash.

However, we and others are developing tooling to view and manage OneTable designs and having a separator mis-match may break such tooling. So we need to fix it.

So, I think we need to push a SemVer major release number change (1.8) and flag this prominently. We could also provide a Table params `legacyUnique` that would preserve the ":" unique separator for those that want to upgrade but not migrate their data.

Lastly, we could provide a Gist snippet for a OneTable Migrate pattern to upgrade unique fields.

mobsense:
The uniqueExpression is not needed as you point out. Expression is only used to parse the properties and prepare the DynamoDB command to be executed. 

removeUnique needs the same treatment.

mobsense:
I'm going to accept this now and then apply these changes to your changes.

Thanks very much for finding this issue and proposing a solution

sdemjanenko:
@mobsense what happens in the case where the value that is supposed to be unique is changed? Can you point me to where that is handled? Thanks!

mobsense:
Update does not currently have that support. Now would be a good time to add.

I'm about to commit changes for:

* Add Table params legacyUniqueSep that will currently default to ':' to preserve current behavior. In version 1.8, that would be removed to take the delimiter value. This will preserve compatibility for the current 1.7 while adopting your fixes for value templates.
* Use the legacyUniqueSep in the unique PK value IF legacyUniqueSep != undefined. If null, then use the delimiter. If set to a character, use that. This is a bridging capability that will not be used in 1.8.
* Added similar changes for removeUnique
* Saved the preparedProperties in params.prepared to tunnel into putItem so we don't do the prepare twice. A small optimization.

mobsense:
Added "early" unique update support and extended update.ts -- more test needed.

PR #107, opened 2021-08-25
Hotfix/additional support type parameter in templates 1

spearmootz:
this is so that the _type field may be used in the sk or pk without any issues


mobsense:
I'm going to close this out. Let me know if we've missed anything in the context fix.

PR #108, opened 2021-08-25
support typeField in templates even if not in properties

spearmootz:


spearmootz:
any updates?

mobsense:
Looks good here and we've been running in production with the mods to addContext for several days.

We'll do an update today with this.

mobsense:
1.7.5 had the context fix. Let me know if that works for you.

spearmootz:
so, 1.7.4 had the issue where sort key was missing.
on 1.7.5 that error is fixed but now my tests fail and tells me i can update variables in my PK, this behavior was not present in 1.7.4

mobsense:
That is really hard to detect when you do a set on a property used in a value template. You can use functions for your value template and then it would be impossible for us to know whether an ingredient of the PK was modified.

If you do it in the properties (first arg), then you are identifying a different item. All OK. When you do it in a set, and that property is used in a value template -- the value used when calculating the value template will be different to the value you set. I don't immediately see a way around that.

mobsense:
I'm going to close this out for now. Please feel free to raise the issue or bug if we can do more.

Issue #115, opened 2021-08-31
[BUG]: .find returns wrong type


spearmootz:
model.Find returns the following type

const model: Paged<Entity<{Schema}>[]>

but Paged already extends an array so it should be 

const model: Paged<Entity<{Schema}>>

mobsense:
Thanks. Looks good.

mobsense:
Thank you for the pull request. Merged and targeted for 1.7.6

spearmootz:
this was actually released in 1.7.5 and looks good on my end

PR #116, opened 2021-08-31
fix .find and .scan types

spearmootz:
This is to fix the following issue.

https://github.com/sensedeep/dynamodb-onetable/issues/115


mobsense:
Looks good. Thank you.

Let us test it out and then approve.

PR #120, opened 2021-09-06
Enhancement: Add `substitutions` to where clauses

rjmackay:
Allow substituting complex variables into where/set clauses using `@{}` brace style.
Substitutions are passed into params.substitutions and extracted into ExpressionAttributeValues.

Implements #119


rjmackay:
Thanks for the quick review!

PR #121, opened 2021-09-07
Fix pre/postFormat typedefs and add Expression type

rjmackay:
- fix: Correct typedefs for preFormat and postFormat
- Add more documentation of postFormat


mobsense:
I'm struggling to see how preFormat can work with being only passed the model. The expression has already parsed fields from the expression this before preFormat is called.

I think preFormat needs to be moved to Model.run and needs to be passed the model and expression. It should be called before invoking expression.command()

Thoughts?

mobsense:
I've moved preFormat to be called from Model.run before calling expression.command.

Also passing the expression which makes it more useful.

Lastly, strengthened  the warning that this is an internal API to the readme.

rjmackay:
> I'm struggling to see how preFormat can work with being only passed the model. The expression has already parsed fields from the expression this before preFormat is called.

I wondered about that so good to fix it. I'll rebase this PR.

rjmackay:
@mobsense I've rebased. I also added types for Expression since you added it to the input for preFormat.
I started with autogenerated types and then filled in as much as I could, it should be good enough for most uses.

dev-embedthis:
That looks clean. Thank you.

When using preFormat, it is really the expression that has all the information.

I would guess that nearly all users should be using postFormat vs preFormat. 

Thanks again, your contributions are really helping.

PR #125, opened 2021-09-10
Fix typing loss when executing array functions on paged

spearmootz:
if you do
const result = await model.find({ some criteria })
you get Paged<Model>
but if you then do
result.(some array function such as filter, map, etc)(someCriteria)
you get any because you lose the typing.

This MR fixes that

### DIFF
--- a/src/Model.d.ts
+++ b/src/Model.d.ts
@@ -170,7 +170,7 @@ export type OneProperties = {
     [key: string]: any
 };
 
-export class Paged<T> extends Array {
+export class Paged<T> extends Array<T> {
     count?: number;
     next?: object;
     prev?: object;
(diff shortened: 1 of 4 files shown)


spearmootz:
i am not sure why the tests failed. should have.

mobsense:
Thanks, Fantastic -- looks great. I'll apply now.

Same comments as on the other PR about more detailed PR descriptions so other folks can fully understand the PR and its scope.

spearmootz:
Hey, just letting you know that this was not released in 1.8.1 but will be in 1.8.2. his was merged before the other PR so just want to make sure that you know that this is unreleased.

mobsense:
Thanks. Just released 1.8.2 now.

PR #126, opened 2021-09-10
mark field as required if required is true.

spearmootz:
So not sure if you guys would like this one but im pretty sure it is best to separate interfaces that came from dynamodb and other places.


mobsense:
Hi, thanks for the PR.

Can you please give a more detail describing the PR.  PR's with a full description are much easier to understand the intent and scope and how it works.

I understand the desire to better handle `required` and you have some pretty mean TS there. Awesome. 

How does this work with `update` where required params are not required in the update if they have already been defined?  This should only apply to `create`.  Is this how you've implemented it?

I don't understand: "it is best to separate interfaces that came from dynamodb and other places"?

Thanks again for your effort.

spearmootz:
so what this does is set that the fully formed model looks like. so think of what dynamodb returns. with required fields as non optional

The EntityAttributes is with everything optional so that you can use it in updates, deletes, and searches.

basically if required is true then the property is not optional, if it is not true or is empty then the property may be undefined.

src/utils.d.ts
```diff
@@ -0,0 +1,5 @@
+type UndefinedProperties<T> = {
```
mobsense, line 1: Can you please add a comment to the file and explain here how this works.

Looks like utter magic to me. To support this, I need to understand how it works.

src/Model.d.ts
```diff
@@ -101,9 +103,14 @@ type EntityField<T extends OneTypedField> =
 /*
     Entities are typed objects whoes signature is based on the schema model of the same name.
  */
-export type Entity<T extends OneTypedModel> = {
-    [P in keyof T]?: EntityField<T[P]>
-};
+export type Entity<T extends OneTypedModel> = UndefinedToOptional<{
+    [P in keyof T]: T[P]['required'] extends true ? EntityField<T[P]> : EntityField<T[P]> | undefined
```
mobsense, line 107: Can you please add a comment to the file and explain here how this works.

spearmootz: DITO.  I wouldnt have been able to create the UndefinedToOptional but i can definitely read it and explain it.

this new extends thing is also new to me, but its basically what you do in the following code, basically if a hardcoded value is of a certain type then the corresponding type can be derived .  I think eventually some more typing should be added for values that have a default value.

```
type EntityField<T extends OneTypedField> =
      T['type'] extends ArrayConstructor ? any[]
    : T['type'] extends BooleanConstructor ? boolean
    : T['type'] extends NumberConstructor ? number
    : T['type'] extends ObjectConstructor ? object
    : T['type'] extends DateConstructor ? Date
    : T['type'] extends StringConstructor ? string
    : T['type'] extends SetConstructor ? Set<T>
    : never;
```

I added comments. to the code

src/utils.d.ts
spearmootz: Similar code to this can be found here. in line 180 utility PickByValue

https://github.com/piotrwitek/utility-types/blob/master/src/mapped-types.ts

src/Model.d.ts
mobsense: The problem for default is that it can be a function. i.e. determined at runtime.

I'm having trouble with the line:

    [P in keyof T]: T[P]['required'] extends true ? EntityField<T[P]> : EntityField<T[P]> | undefined

Is that saying .... If there is a property required that is truthy (extends true), then the result is EntityField<...> otherwise it can be EntityField or undefined.?

src/utils.d.ts
mobsense: That is an awesome resource.

We definitely want zero dependencies. So just a ref here is fine. Is see it is MIT licensed so that is okay to learn from it and adapt. 

Can you add comments to the code to do your best to explain so the code can be supported by others too?

spearmootz: I already did. Let me know if it's not sufficient.

src/Model.d.ts
spearmootz: Yes. Although im not sure if it's truthy or true. In this case it can't be truthy because the field required can only be a boolean

src/utils.d.ts
mobsense: Cool, sorry I missed that. 

Looks great. I'll merge later today.

spearmootz: Absolutely. Thank you!

mobsense:
See issue #185

PR #128, opened 2021-09-14
remove illegal character. ; cannot be on interfaces

spearmootz:
; cannot be in interfaces.

when i changed the interface the other day i left this because i was editing in github.


spearmootz:
node_modules/dynamodb-onetable/dist/mjs/Model.d.ts:56:2 - error TS1036: Statements are not allowed in ambient contexts.

56 };
    ~

PR #151, opened 2021-10-08
DOC: Amend README link to OneTable Migrate

AndrewCathcart:
Noticed the README had a circular link back to this repo when discussing migrations. I believe migration support is achieved via OneTable Migrate & OneTable CLI so I've amended the links. 

Is the "Controller" previously mentioned now the Migration Library?


mobsense:
Thanks @AndrewCathcart.

PR #169, opened 2021-10-29
fix: Fix unique attribute creation when hash and sort key are different

kujtimiihoxha:
It looks like the when creating the item for the unique attribute `onetable` does not use the custom hash and sort keys.


mobsense:
Another good PR. Thank you.

PR #170, opened 2021-10-29
fix: Update model definition

kujtimiihoxha:
The model definition does not have the `schema` option.


mobsense:
Thank you for the PR. I love minimal PRs that are simple and fix things easily! Much appreciated.

PR #171, opened 2021-10-30
Add support for params in group by

kujtimiihoxha:
To be able to use `groupByType` we need to keep the `hidden` fields, there are cases where after we group the items we want those fields to not be present.

This PR adds the ability to provide `OneParams` to `groupByType` with `{parse:true, hidden:false}` so that the items are cleaned before they are added to a group.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ coverage
 test/proto.ts
 test/schemas/protoSchema.ts
 *.keep
+.idea
+yarn.lock
--- a/src/Table.js
+++ b/src/Table.js
@@ -623,10 +623,19 @@ export class Table {
             let type = item[this.typeField] || '_unknown'
             let list = result[type] = result[type] || []
             let model = this.schema.models[type]
-            if (Object.keys(params).length && model && model !== this.schema.uniqueModel) {
-                item = model.transformReadItem('get', item, {}, params)
+            let preparedItem
+            if(typeof params.hidden === 'boolean' && !params.hidden){
+                let fields = model.block.fields
+                preparedItem = {}
+                for (let [name, field] of Object.entries(fields)) {
+                    if (!(field.hidden && params.hidden !== true)) {
+                        preparedItem[name] = item[name]
+                    }
+                }
+            } else {
+                preparedItem = item
             }
-            list.push(item)
+            list.push(preparedItem)
         }
         return result
     }
--- a/src/Table.js
+++ b/src/Table.js
@@ -624,7 +624,7 @@ export class Table {
             let list = result[type] = result[type] || []
             let model = this.schema.models[type]
             let preparedItem
-            if(typeof params.hidden === 'boolean' && !params.hidden){
+            if (typeof params.hidden === 'boolean' && !params.hidden) {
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -6,3 +6,5 @@ coverage
 test/proto.ts
 test/schemas/protoSchema.ts
 *.keep
+.idea
```
kujtimiihoxha, line 9: Added these since they are common  I can remove them if needed.

test/typescript-tenant.ts
```diff
@@ -59,6 +59,19 @@ test('Fetch', async() => {
     expect(collection.User.length).toBe(userData.length)
 })
 
+
+test('Group by with params', async() => {
```
kujtimiihoxha, line 63: I added this test here because I did not know where to add it, please let me know if I should move it.

src/Table.js
```diff
@@ -617,11 +617,15 @@ export class Table {
...
     */
-    groupByType(items) {
+    groupByType(items, params={}) {
         let result = {}
         for (let item of items) {
             let type = item[this.typeField] || '_unknown'
             let list = result[type] = result[type] || []
+            let model = this.schema.models[type]
+            if (Object.keys(params).length && model && model !== this.schema.uniqueModel) {
+                item = model.transformReadItem('get', item, {}, params)
```
mobsense, line 627: I think the items have already been transformed, so this will be calling transformReadItem twice?

kujtimiihoxha: The issue is that to be able to use groupby the items need to have the hidden attributes because it uses _type, I wanted to have a way to remove those hidden attributes after the group by, is there a clear way to do this?

mobsense: Understand.

So having params.hidden == true as you've got works fine.

You then want to copy the iteration logic from transformReadItem to pick out the non-hidden properties.

mobsense: Did you want to have a go at this, or do you want us to do it?

kujtimiihoxha: Hey @mobsense sorry I did not respond, I will get something ready soon, one question I have should I still use the `OneParams` type here or should I just add something like 
```
groupByType(items, removeHidden=false) 
```

This way whoever uses this would know that the other props are not available ?

mobsense: Definitely use a map rather than a single option. That way we can provide for future options.

My preference is to use OneParams. The Model APIs use OneParams and certain APIs only support a sub-set of options as defined by the doc. This is a case where more is less. TS can get very complex where every API has a different params type. So the OneParams is a reasonable medium where the total set of options for APIs are defined.

kujtimiihoxha: Got it 👍

kujtimiihoxha: @mobsense let me know if [this](https://github.com/sensedeep/dynamodb-onetable/pull/171/commits/c474482623468b9da51bd716179fef719e29a25a#diff-a44f85a07e6bd8ef01c1de04a73f6db1da817c9bb48d523071f0f0f63ac20120R627) is what you meant.

mobsense:
Thanks for sticking with this. Appreciated.

PR #180, opened 2021-11-05
fix: Update and remove unique property

kujtimiihoxha:
This is a followup to https://github.com/sensedeep/dynamodb-onetable/pull/169 I did not notice that I had to change multiple places.

### DIFF
--- a/src/Model.js
+++ b/src/Model.js
@@ -604,7 +604,7 @@ export class Model {
         for (let field of fields) {
             let pk = `_unique#${this.name}#${field.attribute}#${properties[field.name]}`
             let sk = `_unique#`
-            await this.schema.uniqueModel.remove({pk, sk}, {transaction})
+            await this.schema.uniqueModel.remove({[this.hash]: pk,[this.sort]: sk}, {transaction})
         }
         await this.deleteItem(properties, params)
         await this.table.transact('write', params.transaction, params)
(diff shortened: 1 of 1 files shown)


mobsense:
Thank you.

trustthe27:
Hi @kujtimiihoxha, thanks for fixing. I had this problem with the 2.1.0 release and it's gone for the current `main` version.

But it seems, there is still a bug left. Because after a delete, the `_unique` item is not removed from the database. So adding another item with the same unique key afterwards will fail.

 That behaviour also doesn't match with the documentation:
 "...
 The created item and the unique item will be created in a transparent transaction so that the item will be created only if all the unique fields are truly unique. The `remove` API will similarly remove the special unique item.
 ..."

trustthe27:
After checking the sourcecode, I know where my problem is rooted.

I tried something like this:

```javascript
  this.UserDbModel.remove(
    { id },
    { transaction },
  );
```

What I needed was something like this, to get the `_unique` item for the field `email` removed as well:

```javascript
this.UserDbModel.remove(
  { id, email },
  { transaction },
);
```

Otherwise the generated PK to delete the `_unique` item is `_unique#User#email#undefined`, what will obviously not work.

For me this works now, but it's an easy trap to fall in 😉

mobsense:
I'll see if we can add a check for unresolved value templates for remove. Should get a warning or exception I think if the keys are not fully specified.

trustthe27:
While experiencing this problem, I would have preferred an exception that kept the data intact, by rolling back the transaction.
Could be done as a check on missing keys, or a DynamoDB condition expression to check if the unique key does exist before removing it.

mobsense:
Just committed a fix. Will throw an exception. Transaction will not complete.

PR #189, opened 2021-11-11
fix: minor cosmetic changes

rdelcampog:
Sorry for monolithic PR but just a few lines changed :)


mobsense:
Thank you. These cosmetic fixups are really valuable to keep the code readable.

PR #192, opened 2021-11-12
Fix ERR_MODULE_NOT_FOUND when loading Error from Expression.js

rjmackay:


mobsense:
Thank you.

I'd love to know which system CAN handle no '.js' and which ones cannot.

rjmackay:
> Thank you.
> 
> I'd love to know which system CAN handle no '.js' and which ones cannot.

Our esbuild+typescript set up handles no extensions on all our own source, but I think it handles things in `node_modules` differently.

rjmackay:
@mobsense actually in this case it might be because I was running through the onetable CLI so it was loading directly with node14 (with ES modules enabled) and no other processing

PR #206, opened 2021-11-24
Handle unique fields that are the same as the prior version

kujtimiihoxha:
If the prior item unique field and the new updated field are the same do not stop updating.

This case happens if you update the whole object not just some fields.


mobsense:
Thank you. Looks good.

PR #209, opened 2021-11-25
Minor readme tweaks regarding groupByType usage

BDQ:
The docs have a few inaccuracies around grouping / fetch:

1) Fetching a collection passing `fetch` an array of types shows the types pluralized in the result (which they are not).

2) `groupByType` requires `hidden: true` be included in the options to ensure `_type` is returned, otherwise the items will all be grouped under a single `_unknown` type.


mobsense:
Thank you for these. Every bit helps. The doc is long and easy to miss things like this. Appreciated.

Issue #220, opened 2021-12-12
TypeScript improvements


markandrus:
Hello 👋 I'm currently evaluating libraries for my application's interface to DynamoDB, and I think OneTable looks great! It's basically what we're doing manually today, but y'all have some additional features (like migration support).

[...]

PR #222, opened 2021-12-13
Improve TypeScript types for find

markandrus:
This addresses #220. I've included two commits:

* **The first makes running `npm test` more robust.** Before this change, for whatever reason, an initial test run would pass, but subsequent runs would fail. There was an issue with DynamoDB being unavailable. Now, both the local Java and Docker paths wait on the port.
* **The second introduces the new types for `find`.** There are probably other types we can improve, but I figure we can tackle those in separate pull requests. I added just a minimal test that the TypeScript types can be used (it's a "list" test that simulates how I use GSIs in my application).

### DIFF
--- a/src/Model.d.ts
+++ b/src/Model.d.ts
@@ -135,6 +135,29 @@ type Entity<T extends OneTypedModel> = Merge<Required<T>, Optional<T>>
  */
 export type EntityParameters<Entity> = Partial<Entity>
 
+export type EntityParametersForFind<T> = {
+    [K in keyof T]: T[K]
+        | Begins<T, K>
+        | BeginsWith<T, K>
+        | Between<T, K>
+        | LessThan<T, K>
+        | LessThanOrEqual<T, K>
+        | Equal<T, K>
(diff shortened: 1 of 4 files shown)


mobsense:
Hi Mark, looks good and thanks for the pickup on the npm test.

I'll merge this, but could you add a few more unit tests to flesh out the test coverage?

mobsense:
HI @markandrus. I've just committed a simple change to add Partial<> to the patch you made.

Please let me know if this impacts any of your behaviour.

    export type EntityParametersForFind<T> = Partial<{...}>

We really need some more units tests for this for TypeScript to catch these errors. Are you able to do that?  Would be wonderful if you can, but if not, let me know and I'll try to resource it.

markandrus:
@mobsense oops, I missed that! The `Partial` fix looks good 👍 I made this change when my team was evaluating dynamodb-onetable, but we had to put that work on pause, so I don't have time to contribute more right now. I hope to when we resume our integration.

PR #237, opened 2022-01-03
Make Schema.version required

icatalina:
Without providing a version the client throws an error

image


mobsense:
Thank you. Good catch.

Issue #245, opened 2022-01-13
Not all fields are returned from an update operation if it executes part of transaction due to unique fields


adrian-grajdeanu:
**Describe the bug**
Update an item without any unique fields, ALL_NEW fields are returned as expected.
However, presence of unique field causes the update to be part of a transaction and in this case not all fields are returned as expected.

Making things more difficult is that even the logging is changed; One gets ```Message: OneTable "transactWrite" "_Generic"```, but no ```Message: OneTable result for "update" "User"```, so what comes back from the db is obscured.

**Cut/Paste**

test/debug.ts
```
import { inspect } from 'util';
import {AWS, Client, Match, Table, Entity, print, dump, delay} from './utils/init'

const Schema = {
  indexes: {
    primary: { hash: 'id' },
  },
  models: {
    User: {
      id: { type: String, },
      nam: {
        type: String,
        unique: true,           // this is the cause of the problem; without this line test succeeds
      },
      cAt: { type: Date, hidden: false },
      uAt: { type: Date, hidden: false },
    },
  },
  version: "0.1.0",
  params: {
    timestamps: true,
    isoDates: true,
    createdField: 'cAt',
    updatedFiled: 'uAt',
  },
  format: "onetable:1.0.0"
}

type UserType = Entity<typeof Schema.models.User>;

const id = 'aswpfhaeswr'
const nam = 'foo'
const nameOther = 'bar'

const table = new Table({
    name: 'DebugTable',
    client: Client,
    schema: Schema,
    logger: (level, message, context) => {
        // if (level == 'trace' || level == 'data') return
        const contextMessage = context ? `Context: ${inspect(context, false, null)}` : ''
        console.log(`At: ${new Date().toISOString()}\nLevel: ${level}\nMessage: ${message}\n${contextMessage}`)
    }
})

test('Create Table', async() => {
    if (!(await table.exists())) {
        await table.createTable()
        expect(await table.exists()).toBe(true)
    }
})

test('Create user', async() => {
    const User = table.getModel('User')
    const user: UserType = await User.create({ id, nam })
    expect(user.nam).toMatch(nam)
    expect(user.cAt).toBeDefined()
})

test('Update nam field should return created field too', async() => {
    const User = table.getModel('User')
    const user: UserType = await User.update({ id, nam: nameOther }, {updateIndexes: true, log: true})
    expect(user.nam).toMatch(nameOther)
    expect(user.cAt).toBeDefined()                     // this is where the problem is observed
[...]

PR #248, opened 2022-01-17
Prevent partial transaction write

kujtimiihoxha:
Remove unique is causing partial transaction execution.

This should fix it and it should behave the same as before for normal use cases.

### DIFF
--- a/src/Model.js
+++ b/src/Model.js
@@ -613,6 +613,7 @@ export class Model {
         Remove an item with unique properties. Use transactions to remove unique items.
     */
     async removeUnique(properties, params) {
+        let transactHere = params.transaction ? false : true;
         let transaction = params.transaction = params.transaction || {}
         let {hash, sort} = this.indexes.primary
         let fields = Object.values(this.block.fields).filter(f => f.unique && f.attribute != hash && f.attribute != sort)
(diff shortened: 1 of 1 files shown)


src/Model.js
```diff
@@ -628,7 +629,10 @@ export class Model {
             await this.schema.uniqueModel.remove({[this.hash]: pk,[this.sort]: sk}, {transaction})
         }
         await this.deleteItem(properties, params)
-        await this.table.transact('write', params.transaction, params)
+        // Only execute transaction if we are not in a transaction
+        if(transactHere){
```
mobsense, line 633: Small nit. Formatting. Space before the {

mobsense:
Thank you. Looks good.

kujtimiihoxha:
@mobsense Done :)

mobsense:
Thank you.

kujtimiihoxha:
@mobsense when is the next release scheduled ? this is a bit blocking for me 😅

mobsense:
We're just mulling #245. As this is going to be a feature release 2.3, this is the opportunity to address it as it will be a breaking change.

kujtimiihoxha:
@mobsense thanks, I got a temporary solution using my fork and a branch as a package dependency so this is not urgent anymore 👍

PR #251, opened 2022-01-17
fix: timestamps + exists: null causes set to be overwritten

onhate:
When `timestamps: true` running an update with `set` and  `exists: null` causes the `set` values to be overwritten by the set of created at property.

image

### DIFF
--- a/src/Model.js
+++ b/src/Model.js
@@ -811,7 +811,10 @@ export class Model {
             if (params.exists == null) {
                 let field = this.block.fields[this.createdField] || this.table
                 let when = (field.isoDates) ? now.toISOString() : now.getTime()
-                params.set = { [this.createdField]: `if_not_exists(\${${this.createdField}}, {${when}})` }
+                params.set = {
+                    ...params.set,
+                    [this.createdField]: `if_not_exists(\${${this.createdField}}, {${when}})`
+                }
             }
         }
         properties = this.prepareProperties('update', properties, params)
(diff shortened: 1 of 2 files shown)


src/Model.js
```diff
@@ -811,7 +811,10 @@ export class Model {
             if (params.exists == null) {
                 let field = this.block.fields[this.createdField] || this.table
                 let when = (field.isoDates) ? now.toISOString() : now.getTime()
-                params.set = { [this.createdField]: `if_not_exists(\${${this.createdField}}, {${when}})` }
+                params.set = {
+                    ...params.set,
+                    [this.createdField]: `if_not_exists(\${${this.createdField}}, {${when}})`
+                }
```
mobsense, line 817: Could this not be more easily written as:

params.set = params.set || {}
params.set[this.createdField] = `if_not_exists(\${${this.createdField}}, {${when}})`

test/timestamps.ts
```diff
@@ -0,0 +1,63 @@
...
+    user = await User.update({ id }, { exists: null, set: { name: 'Marcelo' }})
+    expect(user.name).toEqual('Marcelo')
+    expect(user.createdAt).toEqual(createdAt)
+    expect(user.updatedAt.getTime()).toBeGreaterThan(updatedAt.getTime())
+})
+
+test('Destroy Table', async() => {
+    await table.deleteTable('DeleteTableForever')
+    expect(await table.exists()).toBe(false)
+})
```
mobsense, line 63: Extra tests ... Love it.

mobsense:
Thank you for this. Looks right to me. Let me test more and then approve. Have a look at the simplification comment I made.

src/Model.js
onhate: yeah, looks better indeed, will apply it

onhate:
@mobsense do you have and idea of when will this fix be released?

mobsense:
Getting close, hopefully a few days. Waiting for feedback on #245

PR #252, opened 2022-01-18
Correct and Adjust a syntax error in Javascript example in README

Sparkboxx:
The provided example in the README shows how to apply a ```list_append``` into ```update.set``` operation. The old example incorrectly uses the ```+``` between arguments. That needed replacement with a ```,```. 

Furthermore we adjusted the example with a ```if_not_exists``` helper to be able to set the initial contact.


mobsense:
Thank you for this quick fix. Little improvements like this really help to drive up quality. Thank you.

PR #258, opened 2022-01-22
feat: DataLoader for get request

onhate:
Original discussion https://github.com/sensedeep/dynamodb-onetable/discussions/250

In this PR I added the https://www.npmjs.com/package/dataloader dependency that allows to group all get requests made in parallel and perform a single BatchGetRequest.

For example (see tests for more) 

```js
// this will accumulate all gets request into a single batch request
// and return an array of items
const items = await Promise.all(
    ids.map(id => model.get({ id })
);
```

```js
// this will perform a single batch get operation with just one item get and will return to each get the same value
// dynamo does not supports duplicate keys on batch request and the implementation will deal with that
const items = await Promise.all([
  model.get({ id: 'same-id' }),
  model.get({ id: 'same-id' }),
  model.get({ id: 'same-id' })
]);
```

it solves this discussion https://github.com/sensedeep/dynamodb-onetable/discussions/235
```js
// this will perform 2 batch requests one with 25 items and another with the extra 5
// as DataLoader is configured with maxBatchSize = 25
// dynamo has a limit of 25 items on batch request
const items = await Promise.all([
  model.get({ id: '1' }),
  model.get({ id: '2' }),
  ...
  model.get({ id: '30' })
]);
```

With this implementation `batchGet` gets deprecated as all `get` request will already be grouped and turned into `batchRequest` internally.

P.S.: there was lots of lint errors and I fixed all of them on this PR too, the implementation is all on `Table.js` and the other files changes are just lint fixes

~TODO: add config to make this feature optional and fallback to original implementation~

### DIFF
--- a/src/Table.js
+++ b/src/Table.js
@@ -5,12 +5,14 @@
  */
 
 import Crypto from 'crypto'
-import UUID from './UUID.js'
+import DataLoader from 'dataloader'
+import { OneTableArgError, OneTableError } from './Error.js'
+import { Expression } from './Expression.js'
+import { groupByMapping } from './helpers'
+import { Metrics } from './Metrics.js'
+import { Schema } from './Schema.js'
 import ULID from './ULID.js'
-import {Expression} from './Expression.js'
-import {Schema} from './Schema.js'
-import {Metrics} from './Metrics.js'
-import {OneTableError, OneTableArgError} from './Error.js'
+import UUID from './UUID.js'
 
 /*
     AWS V2 DocumentClient methods
@@ -25,7 +27,7 @@ const DocumentClientMethods = {
     batchGet: 'batchGet',
     batchWrite: 'batchWrite',
     transactGet: 'transactGet',
-    transactWrite: 'transactWrite',
+    transactWrite: 'transactWrite'
 }
 
 /*
@@ -48,10 +50,12 @@ const DynamoOps = {
     batchGet: 'batchGet',
     batchWrite: 'batchWrite',
     transactGet: 'transactGet',
-    transactWrite: 'transactWrite',
+    transactWrite: 'transactWrite'
 }
 
 const GenericModel = '_Generic'
+
+const maxBatchSize = 25
 
 /*
     Represent a single DynamoDB table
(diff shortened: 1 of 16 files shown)


src/Table.js
```diff
@@ -519,42 +524,49 @@ export class Table {
...
             if (params.stats || this.metrics || this.monitor) {
                 cmd.ReturnConsumedCapacity = params.capacity || 'INDEXES'
                 cmd.ReturnItemCollectionMetrics = 'SIZE'
             }
-            this.log[params.log ? 'info' : 'trace'](`OneTable "${op}" "${model}"`, {trace})
-            if (this.V3) {
-                result = await client[op](cmd)
+            if (op === 'get') {
+                result = await this.loader.load(cmd)
             } else {
```
onhate, line 537: here is the main change, instead of directly executing the get operation it delegates the get to the loader that will accumulate the gets and perform a single batch get.

src/Table.js
```diff
@@ -697,27 +709,88 @@ export class Table {
...
+    /**
+     * this is a function passed to the DataLoader that given an array of Get commands
+     * will group all of them by TableName and make all Keys unique and then will perform
+     * a BatchGet request and process the response of the BatchRequest to return an array
+     * with the corresponding response in the same order as it was requested.
+     *
+     * @param commands
+     * @returns {Promise<*>}
+     */
+    async groupByTableAndBatchGet(commands) {
```
onhate, line 738: this is the method that DataLoader will use to batch get all items, it receives an array of Get commands and return an array of `{ Item }` so the get command continues thinking it was just a regular Get

mobsense:
Thank you. Really appreciated. 

Can I give some feedback to help us digest this.

There is lots of code reformatting which makes reading the diffs difficult and doing a merge/pull risky. Can you resubmit the pull request without any code reformatting?

Can you also please resubmit the pull request in 2 parts: the data loader patch and the lint fixes. Both without any code reformatting. That way we can unlink the change sets and make each pull request focussed on one thing. We need a pull request to be minimal so we can properly assess every line change. We are very careful about this.

Second, we want zero external dependencies. This is for risk/security as well as minimal footprint. So can you use pass a data loader instance in via the Table constructor. That way, users who won't use this, don't pull in the data loader dep. Users who do want it, can construct the data loader when they create the table.

PR #259, opened 2022-01-23
feat: DataLoader

onhate:
Original discussion https://github.com/sensedeep/dynamodb-onetable/discussions/250

In this PR I added the possibility to use https://www.npmjs.com/package/dataloader which allows to group all get requests made in parallel and perform a single BatchGetRequest.

For example (see tests for more) 

```js
// this will accumulate all gets request into a single batch request
// and return an array of items
const items = await Promise.all(
    ids.map(id => model.load({ id })
);
```

```js
// this will perform a single batch get operation with just one item get and will return to each get the same value
// dynamo does not supports duplicate keys on batch request and the implementation will deal with that
const items = await Promise.all([
  model.load({ id: 'same-id' }),
  model.load({ id: 'same-id' }),
  model.load({ id: 'same-id' })
]);
```

it solves this discussion https://github.com/sensedeep/dynamodb-onetable/discussions/235
```js
// this will perform 2 batch requests one with 25 items and another with the extra 5
// as DataLoader is configured with maxBatchSize = 25
// dynamo has a limit of 25 items on batch request
const items = await Promise.all([
  model.load({ id: '1' }),
  model.load({ id: '2' }),
  ...
  model.load({ id: '30' })
]);
```

With this implementation `batchGet` kind of gets deprecated as all `load` request will already be grouped and turned into `batchRequest` internally, if later we change the `batch` param on `get` to boolean and use internally the `load` method when it's true.

For example:

```js
async get(properties = {}, params = {}) {
    if (params.batch) {
        return await this.load(properties, params)
    }
    // do regular get
 }
```

That would be a great value to not have to use batch object and call batchGet manually.

### DIFF
--- a/test/loader.ts
+++ b/test/loader.ts
@@ -1,0 +1,85 @@
+/*
+   Batch loader
+ */
+
+// @ts-ignore
+import DataLoader from 'dataloader';
+import { DefaultSchema } from './schemas';
+import { Client, dynamoExecutedCommandsTracer, Table } from './utils/init';
+
+const table = new Table({
+  name: 'BatchTest',
+  client: Client,
+  schema: DefaultSchema,
+  dataloader: DataLoader
+});
+
+let users: any[];
+
+let data = [
+  { name: 'Peter Smith', email: 'peter@example.com', status: 'active' },
+  { name: 'Patty O\'Furniture', email: 'patty@example.com', status: 'active' },
+  { name: 'Cu Later', email: 'cu@example.com', status: 'inactive' }
+];
+
+describe('Loader', () => {
+  test('Create', async () => {
+    if (!(await table.exists())) {
+      await table.createTable();
+    }
+  });
+
+  test('Seed data', async () => {
+    const batch = {};
+    for (let item of data) {
+      await table.create('User', item, { batch });
+    }
+    await table.batchWrite(batch);
+    users = await table.scan('User');
+  });
+
+  test('Batch load', async () => {
+    // this will accumulate all gets request into a single batch request
+    const items = await Promise.all(
+      users.map(user => {
+        return table.load('User', { id: user.id });
+      })
+    );
+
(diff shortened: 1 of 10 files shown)


onhate:
@mobsense here it is :)

test/utils/init.ts
```diff
@@ -13,6 +15,12 @@ const Client = new Dynamo({
             accessKeyId: 'test',
             secretAccessKey: 'test',
         }),
+        logger: {
```
mobsense, line 18: I'm not familiar with this param to the DynamoDBClient. What is dynamoExecutedCommandsTracer doing for you?

test/loader.ts
```diff
@@ -0,0 +1,85 @@
...
+   Batch loader
+ */
+
+// @ts-ignore
+import DataLoader from 'dataloader';
+import { DefaultSchema } from './schemas';
+import { Client, dynamoExecutedCommandsTracer, Table } from './utils/init';
+
+const table = new Table({
+  name: 'BatchTest',
```
mobsense, line 11: Small nit. Can you format all your code / files with TS=4 spaces

mobsense: I'm just picking at little things to tidy up before merging.

I'd like to keep this under a prototype status in the main branch. i.e. Not document yet while you battle harden it in practice and others get to play with it. Once documented, it is set in stone.

test/utils/init.ts
onhate: This is used to track all the commands performed on dynamodb, if you see the tests on `loader.ts` I check the following 

```js
 expect(dynamoExecutedCommandsTracer).toHaveBeenCalledWith(
      expect.objectContaining({
        commandName: 'BatchGetItemCommand'
      })
    );
    ```

test/loader.ts
onhate: done the reformatting, ready to go I suppose :)

mobsense:
Thanks for your patience working the small issues. Much appreciated.

Issue #255, opened 2022-01-19
Transform object for read after write operations


mobsense:
### Discussed in https://github.com/sensedeep/dynamodb-onetable/discussions/254

<div type='discussions-op-text'>

<sup>Originally posted by **onhate** January 19, 2022</sup>
Hi, due to the limit on the record size I implemented a compressed field on my table , it should behave similar to crypt logic but compress on write and uncompress on reads, for that I used the table transform method with following logic:

```
transform(model: AnyModel, op: string, item: AnyEntity) {
  if (op === 'read') {
    if (item.content) {
      item.content = 'uncompressed content';
    }
  }
  if (op === 'write') {
    if (item.content) {
       item.content = 'compressed content';
    }
  }
  return item;
},
```

But I discovered that when I call `await model.create({  content: 'Very Large String' });` the returned value of the create method is the value parsed as `write` which means it returns the object with compressed string value, to circumvent that I need to perform a `get` after `create` so it returns the uncompressed value.

That makes me think if the returned value of `create`/`update` should also be parsed as read after executing the operation on database.

WDYT?
</div>

PR #261, opened 2022-01-24
fix: call table.transform as read after write

onhate:
Closes https://github.com/sensedeep/dynamodb-onetable/issues/255

### DIFF
--- a/src/Model.js
+++ b/src/Model.js
@@ -18,6 +18,15 @@ const ReadWrite = {
     update: 'write'
 }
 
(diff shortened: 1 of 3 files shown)


mobsense:
Thank you.

onhate:
@mobsense could I get a minor release with this fix? :)

mobsense:
Sure, I'm just waiting for feedback on the 2.3 release to make sure we've not overlooked anything.

Perhaps Sunday.

PR #268, opened 2022-02-04
fix Set type

onhate:
for this schema 

```
const ReactionsSchema = {
  fields: {
    likes: { type: Set },
    dislikes: { type: Set },
  }
}
```

when doing 

 ```
 type Reactions = Entity<typeof ReactionsSchema.fields>;
```

the type of `Reactions.like` and `Reactions.dislikes` are resolved as `Set<{ type: Set }>`


mobsense:
Thank you!

PR #270, opened 2022-02-05
fix: dataloader keys manipulation logic when v2 vs v3

onhate:
- fixed logic for dataloader keys manipulation when client v2 vs v3 (v3 has `{S:''}` and v2 is `''`)
- run all tests both with v2 and v3

### DIFF
--- a/src/Table.js
+++ b/src/Table.js
@@ -7,10 +7,10 @@
 import Crypto from 'crypto'
 import UUID from './UUID.js'
 import ULID from './ULID.js'
-import { Expression } from './Expression.js'
-import { Schema } from './Schema.js'
-import { Metrics } from './Metrics.js'
-import { OneTableArgError, OneTableError } from './Error.js'
+import {Expression} from './Expression.js'
+import {Schema} from './Schema.js'
+import {Metrics} from './Metrics.js'
+import {OneTableArgError, OneTableError} from './Error.js'
 
 /*
     AWS V2 DocumentClient methods
@@ -84,7 +84,7 @@ export class Table {
         this.setParams(params)
         this.schema = new Schema(this, params.schema)
         if (params.dataloader) {
-            this.dataloader = new params.dataloader(cmds => this.batchLoaderFunction(cmds), { maxBatchSize })
+            this.dataloader = new params.dataloader(cmds => this.batchLoaderFunction(cmds), {maxBatchSize})
         }
     }
 
@@ -774,16 +774,16 @@ export class Table {
                 })
                 return index2 === index1
             })
-            requestItems[tableName] = { Keys: uniqueKeys }
+            requestItems[tableName] = {Keys: uniqueKeys}
             return requestItems
         }, {})
 
-        const results = await this.batchGet({ RequestItems: requestItems })
+        const results = await this.batchGet({RequestItems: requestItems})
 
         // return the exact mapping (on same order as input) of each get command request to the result from database
         // to do that we need to find in the Responses object the item that was request and return it in the same position
         return commands.map((command, index) => {
-            const { model, params } = expressions[index]
+            const {model, params} = expressions[index]
 
             // each key is { pk: { S: "XX" } } when V3 or { pk: "XX" } when V2
             // on map function, key will be pk and unmarshalled will be { S: "XX" }, OR "XXX"
--- a/src/Table.js
+++ b/src/Table.js
@@ -7,10 +7,10 @@
 import Crypto from 'crypto'
 import UUID from './UUID.js'
 import ULID from './ULID.js'
-import {Expression} from './Expression.js'
-import {Schema} from './Schema.js'
-import {Metrics} from './Metrics.js'
-import {OneTableError, OneTableArgError} from './Error.js'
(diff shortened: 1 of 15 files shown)


test/crypto.ts
```diff
@@ -50,10 +50,18 @@ test('Get', async() => {
 
 test('Get raw ', async() => {
     user = await User.get({id: user.id}, {hidden: true, parse: false})
-    expect(user.pk.S).toMatch(/^User#/)
-    expect(user.email.S).toBeDefined()
-    expect(user.email.S).not.toMatch('peter@example.com')
-    expect(user.email.S).toMatch(/^primary/)
+    if (isV3()) {
```
mobsense, line 53: User.get should not ever be returning email.S.  Should these values not be unmarshalled into just a pure email?

test/unique.ts
```diff
@@ -39,8 +39,16 @@ test('Create user 1', async() => {
     let items = await table.scanItems()
     expect(items.length).toBe(3)
 
-    let unique = items.filter((item) => item.pk.S.indexOf('interpolated') >= 0)
-    let pk = unique[0].pk.S
+    let pk = (() => {
+        if (isV3()) {
+            let unique = items.filter((item) => (item.pk.S).indexOf('interpolated') >= 0)
+            return unique[0].pk.S
```
mobsense, line 45: Same as per previous comment. V3 and V2 should be returning identical data.

src/Table.js
```diff
@@ -7,10 +7,10 @@
 import Crypto from 'crypto'
 import UUID from './UUID.js'
 import ULID from './ULID.js'
-import {Expression} from './Expression.js'
-import {Schema} from './Schema.js'
-import {Metrics} from './Metrics.js'
-import {OneTableError, OneTableArgError} from './Error.js'
+import { Expression } from './Expression.js'
```
mobsense, line 10: Formatting.

mobsense:
Thanks for the PR.

General question ... should not V2 and V3 be returning identical results. We don't want to be surfacing V2/V3 to the user (in general) and the loader should be returning identical results regardless of the SDK.

onhate:
Hey, hi @mobsense, the loader is not returning different values, I just enabled a way to run all tests both on v2 and v3 and some unrelated tests with dataloader failed, all of them when `parse: false` which I would consider fair as v2 uses document client that already unmarshall data

mobsense:
Can you help me understand the case with dataloader failing and needing parse: false?  In general, I've tried to avoid V2/V3 percolating up even in tests. Especially for V3 as that will become the dominant SDK and, again in general, we don't want to be surfacing the underlying DynamoDB encoding unmarshalled format.

onhate:
The daloader does not needs parse as false, what happens is that as it uses the generated expressions to map the response from dynamodb and different clients generates different expressions and responses

onhate:
If you check the dataloader test you will see the response is always parsed correctly both on v2 and v3, the difference on tests from v2 and v3 only appears when asking for unparsed values, which is correct in guess

onhate:
@mobsense here is the flow and why loader needs to deal with ddb client specifics:

1. model.load is called
2. data loader accumulates the expressions to batch get
3. expressions can be either v2 (unmarshalled) or v3 (marshalled)
4. batch get does not supports duplicate keys on batch get so we need to "cleanup" the array of expressions
5. as expressions can be either v2 or v3 we need (internally only) check if what is the version used to make the expressions unique https://github.com/sensedeep/dynamodb-onetable/pull/270/files#diff-a44f85a07e6bd8ef01c1de04a73f6db1da817c9bb48d523071f0f0f63ac20120R768
6. after batch get is run we need now to return an array of items matching the exact same position they were requested

for example: 

if you requested `[{pk:1, sk:1}, {pk:1,:sk1},{pk2:sk:2}]` I run a batch get only for `[{pk:1, sk:1}, {pk2:sk:2}]` but the response required 3 items so I need to loop the responses (that can be either v2 or v3) and match the key with the item to return 3 items matching the exact same order they were requested.

In summary, internally we need to know if it is v2 or v3 to make keys unique and process response but the output of the dataloader is exactly the same as if it was a `get` call


Why differences for v2 and v3 on tests?

Well, I noticed we were not fully testing v2 so I made a small change on tests to run all suites both with v2 and v3 and when I did that some unrelated with dataloader tests, all using `parse: false` failed because they were only written for v3. In my opinion if you ask for unparsed values and we return raw values from dynamodb it's expected that the response will vary depending on the client you use.

test/crypto.ts
onhate: as explained here https://github.com/sensedeep/dynamodb-onetable/pull/270#issuecomment-1030811573 it is because this test asks for `parsed: false` values

test/unique.ts
onhate: https://github.com/sensedeep/dynamodb-onetable/pull/270#discussion_r800164728

mobsense:
Looks good. Just need the conflict with index.js resolved to merge.

onhate:
fixed, thanks @mobsense

mobsense:
Great, thank you!

PR #271, opened 2022-02-06
fix module import

shishkin:
I ran into an issue when using OneTable in ESM setup:

```
/.../node_modules/ts-node/dist-raw/node-esm-resolve-implementation.js:383
    throw new ERR_MODULE_NOT_FOUND(
          ^
CustomError: Cannot find module '/.../node_modules/dynamodb-onetable/dist/mjs/Error' imported from /.../node_modules/dynamodb-onetable/dist/mjs/index.js
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -5,7 +5,7 @@
 import { Expression } from './Expression.js'
 import { Model } from './Model.js'
 import { Table } from './Table.js'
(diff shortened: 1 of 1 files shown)


mobsense:
Thank you. I wish these issues would be solved on all platforms once and for all. ESM support in Node is one big bed of nails.

PR #274, opened 2022-02-09
Handle error narrowing with aws-sdk v3

dibell:
aws-sdk v3 seems to have changed where the error types are stored. Previously it was in .code. Now it is in .name. See https://github.com/aws/aws-sdk-js-v3/issues/2874

The consequence of that is that error narrowing no longer works. E.g. creating an item twice does not raise a neat ConditionError any more. It is a generic failure that is harder to handle.

This PR aims to resolve that issue while retaining compatibility with v2.

### DIFF
--- a/src/Error.js
+++ b/src/Error.js
@@ -22,8 +22,9 @@ function init(self, message, context) {
 
 export function checkErr(error, code) {
     // aws sdk v3 has moved code to name - see https://github.com/aws/aws-sdk-js-v3/issues/2874
-    return error.code == code || error.name == code; 
+    return error.code == code || error.name == code
 }
+
 export class OneTableError extends Error {
     constructor(message, context) {
         super(message)
--- a/src/Error.js
+++ b/src/Error.js
@@ -20,6 +20,10 @@ function init(self, message, context) {
     }
 }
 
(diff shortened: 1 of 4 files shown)


mobsense:
All I can say is UGH!  Why do they do this?

I'd like to take a bit of time to see if we could alias err.code = err.name internally and not have this leak out of the library. We do catch AWS exceptions in most places (not all) and we could set err.code before re-throwing. That way the error exception object could always contain an `err.code`. 

Thoughts?

dibell:
Doesn't the error usually get wrapped in a OneTableError?

mobsense:
If it does get full wrapped, then we don't want to surface that outside the OneTable API. There may be some cases where we do not always catch the errors.

I'd rather go down the path of internally in Error.js storing the error always in err.code and handling the V2 / V3 differences inside Error.js.

mobsense:
So we've added the V3 conversion from err.name to err.code and also tightened to capture all AWS exceptions in execute().

Does this address all your concerns?  i.e. are there things in this pull request that still need addressing.

dibell:
The changes in main cover the err.name handling nicely. Many thanks.

PR #279, opened 2022-02-17
Add Table.child typescript definition

smacker:
Table object has a very useful method `child` but typescript definition for it is missing.


mobsense:
Thank you

PR #281, opened 2022-02-18
Fix Model.update with unknown field and unique schema

smacker:
Currently for such a use case (see the test) exception is raised.

```
TypeError: Cannot read properties of undefined (reading 'unique')
```


test/unique.ts
```diff
@@ -79,13 +79,15 @@ test('Update user 2 with unique email', async() => {
...
+test('Update with unknown property', async() => {
     const props = {
         name: 'Judy Smith',
-        age: 42,
+        age: 15,
+        unknown: 'value',
     }
     user = await User.update(props, {return: 'get'})
-    expect(user).toMatchObject(props)
+    const {unknown, ...expectedProps} = props;
```
mobsense, line 89: semicolon

test/unique.ts
```diff
@@ -79,13 +79,15 @@ test('Update user 2 with unique email', async() => {
 })
 
 
-test('Update non-unique property', async() => {
```
mobsense, line 82: Do we still want the old non-unique property test case. i.e. should we add a new test case rather than replace the old one?

test/unique.ts
smacker: fixed

mobsense:
Thank you!

PR #282, opened 2022-02-18
Fix null property deref in OneTableArgError

tripplilley:
`context` argument is allowed to be null or undefined by type definition, and many callers (e.g., Table.js)
use the `message`-only constructor, so the deref of `.code` needs to use the null/undef-safe operator `?.`

Discovered by feeding a "works with an older version" schema to the `Table` constructor and getting
```
TypeError: Cannot read property 'code' of undefined
    at new OneTableArgError
```

### DIFF
--- a/src/Error.js
+++ b/src/Error.js
@@ -44,6 +44,6 @@ export class OneTableArgError extends Error {
     constructor(message, context) {
         super(message, context)
         init(this, message, context)
-        this.code = context.code || 'ArgumentError'
+        this.code = context?.code || 'ArgumentError'
     }
 }


mobsense:
Thanks for spotting this.

I think there are a few other paths require context to be non-null. So I'd propose we do:

```
export class OneTableError extends Error {
    constructor(message, context = {}) {
```

That would then ensure that this.context references are always valid.

mobsense:
I've applied the proposed change. Do we still need parts of this pull request or is it fully addressed now?

mobsense:
Closing for now, please feel free to resubmit

tripplilley:
> I've applied the proposed change. Do we still need parts of this pull request or is it fully addressed now?

No, this addresses my issue. Sorry for the delayed reply. Looks good, thanks!

PR #283, opened 2022-02-20
Add missing check for numeric value

gregplaysguitar:
The problem referenced in the comment:

> Except for numbers because we don't want to confuse valuesMap indexes. i.e. 7 vs "7" 

was still present if a number came first in the attribute list, i.e. `{"version": 1, "id": "1"}` - in this case you get id set to the number `1` instead of a string.


mobsense:
Thank you.

PR #284, opened 2022-02-21
fix: Allow substitutions to be falsy

rjmackay:
Missing substitutions was being thrown if the substitution was supplied but falsy. This is particularly problematic when filtering on a boolean value.


mobsense:
Thanks Robbie.

rjmackay:
> Thanks Robbie.

In this case I'm pretty sure this was my own code to begin with, so I'm just fixing my own bugs really 😂

PR #287, opened 2022-02-21
feat: push param shortcut

onhate:
A shortcut on params to allow safe pushing to an array using `set ${target} = list_append(if_not_exists(${target}, ${empty}), ${items})`

This is a [very common approach](https://medium.com/@nehaguptag/append-to-dynamodb-list-golang-5ac7b3ec5559) on DynamoDB and I see myself repeating it many times in my codebase

From

```js
await User.update(
   { id: user.id },
   {
     set: { tokens: 'list_append(if_not_exists(${tokens}, @{empty}), @{token})' },
     substitutions: { token: ["yellow"], empty: [] }
   }
);
```

To

```js
await User.update(
   { id: user.id },
   { push: { tokens:  ["yellow"] }}
)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1393,14 +1393,20 @@ The method returns the all the Javascript properties for the item. Hidden attrib
 
 If the method fails to update, it will throw an exception. If `params.throw` is set to false, an exception will not be thrown and the method will return `undefined`.
 
-The optional params are described in [Model API Params](#model-api-params).    
+The optional params are described in [Model API Params](#model-api-params).
+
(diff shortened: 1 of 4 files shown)


mobsense:
Thanks for the PR. Will take a few days before I can get to this. Will get back to you next week on it.

mobsense:
This patch being larger is taking a bit of time to check. We don't have use cases that exercise this code, so we're being a bit more cautious. No issues, just caution.

onhate:
Got it, let me know if you find something to be changed

PR #290, opened 2022-02-22
fix: ulid validation regex in example

hicksy:
I noticed that the `validation` regex was incorrect for the generated `ulid` field.


mobsense:
Thanks. Always good to have correct doc.

PR #308, opened 2022-03-25
Make better use of type inference

thdxr:
I think there's a lot of typesafety + autocomplete support we can add to this library! I did a quick initial pass to demonstrate an example - if this looks good I can keep going through everything I find.

This PR adds inference to the `getModel` function

Before:
```
const model = table.getModel<TestModel>("Test")
```

After:
```
const model = table.getModel("Test")
```

The latter gives you autocomplete on the string and also gives the correct return type

2022-03-24-205506_836x446_scrot

Also is there any chance you'd add typescript directly to this project so most of the definition files can be generated instead of manually maintained?

### DIFF
--- a/src/Table.d.ts
+++ b/src/Table.d.ts
@@ -8,7 +8,7 @@ export type EntityGroup = {
     [key: string]: AnyEntity[]
 };
 
-type TableConstructorParams = {
+type TableConstructorParams<Schema extends OneSchema> = {
     client?: {},                    //  Instance of DocumentClient or Dynamo.
     crypto?: {},                    //  Crypto configuration.
     generate?: (() => string),      //  Function to generate IDs for field schema that requires.
@@ -18,7 +18,7 @@ type TableConstructorParams = {
     intercept?: (model: AnyModel, op: string, rec: {}, params: OneParams, raw?: {}) => void,
     metrics?: boolean | object,     //  Enable CloudWatch metrics.
     name?: string,                  //  Table name.
-    schema?: OneSchema,             //  Table models schema.
+    schema?: Schema,             //  Table models schema.
     senselogs?: {},                 //  SenseLogs instance for logging
     //  Transform record for read / write.
     transform?: (model: AnyModel, op: string, item: AnyEntity, properties: OneProperties, params?: OneParams, raw?: {}) => AnyEntity,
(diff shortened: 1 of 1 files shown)


mobsense:
This looks really good. Thank you ... keep going !!!

rjmackay:
❗ 😁 Happy to see TS support improving!

mobsense:
We could convert the library to TS. However, I'm being deliberately slow so we focus first on functionality and the core API.

I've seen too many projects that have converted to TS and the codebase ends up being destroyed in overly complex TS syntax. You end up with code that is pretty obscure and hard to maintain and hack.

I think we are getting close to being able to do this. But I'm going to be firm on getting a MINIMAL TS definition rather than trying to cover every little TS type issue. Perhaps in a month or two, we can start this.

Who is interested in helping?

mobsense:
Hi @thdxr 

Can you check into why the integration tests are failing and remedy?

Thanks

mobsense:
Just following up on this. The pull request looks interesting, just need the unit tests to be fixed to apply.

cdash04:
It would be great to add a return type to marshall and unmarshall while making changes in `Table.d.ts`:

```typescript
marshall(item: AnyEntity | AnyEntity[], params?: OneParams): AnyEntity;
unmarshall(item: AnyEntity | AnyEntity[], params?: OneParams): AnyEntity;
```

mobsense:
Just added that change for marshall to the repo. Thanks.

mobsense:
Anyone want to dig in and get these unit tests working so we can merge?

loetjvr:
Hi, I believe I got this working but unsure on 2 things.

1. Not sure how to assert typescript other than hovering over it in the editor. Should probably have a type check step right?
2. How do I update this specific PR? or do I just create a new one?

mobsense:
Feel free to create a new PR. That may be simpler.

I'm not sure I've got an answer for (1). I'm not a TS expert myself. But that would definitely be very worthwhile.

rjmackay:
> 1. Not sure how to assert typescript other than hovering over it in the editor. Should probably have a type check step right?

You can usually run `tsc` and it'll shout about type errors (add `--noEmit` to skip the build output)

thdxr:
Going to close this PR as I'm not using this library anymore

mobsense:
Okay, can you give some feedback as to what you are using and what we can improve?

thdxr:
Sorry realizing that sounded rude. I moved to ElectroDB primarily because typesafety + inference is a first class feature and those are things I particularly care about.

mobsense:
Now worries. Thanks for the feedback.

Soon, when we reach a stable plateau, we'll convert the code to a TS core. That should fill in the gaps in our TS coverage.

PR #313, opened 2022-03-31
Return OneTableErrors with code:UniqueError when values are not unique

rjmackay:
This was failing to properly handle errors and just returning the TransactionCancelled error

### DIFF
--- a/src/Model.js
+++ b/src/Model.js
@@ -545,10 +545,10 @@ export class Model {
         try {
             await this.table.transact('write', params.transaction, params)
         } catch (err) {
-            if (err.message.indexOf('ConditionalCheckFailed') >= 0) {
+            if (err instanceof OneTableError && err.code === 'TransactionCanceledException' && err.context.err.message.indexOf('ConditionalCheckFailed') !== -1) {
                 let names = fields.map(f => f.name).join(', ')
                 throw new OneTableError(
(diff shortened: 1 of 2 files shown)


rjmackay:
Ah it looks like I missed some failures with dynamo v2. I'll take a look at those soon, any pointers on that would be appreciated.

src/Model.js
```diff
@@ -545,10 +545,10 @@ export class Model {
         try {
             await this.table.transact('write', params.transaction, params)
         } catch (err) {
-            if (err.message.indexOf('ConditionalCheckFailed') >= 0) {
+            if (err instanceof OneTableError && err.code === 'TransactionCanceledException' && err.context.err.CancellationReasons.map(r => r.Code).includes('ConditionalCheckFailed')) {
```
rjmackay, line 548: Does this need to be more blunt to work with dynamodb v2? ie. not checking CancellationReasons

mobsense: Sorry, I don't have immediate understanding of what the V2 equivalent would be.

I do seem to remember that V2 did not include such transaction failure errors. The GO SDK did and there was a message somewhere that said V2 would not support it in Javascript.

rjmackay:
Updated and fixed tests

mobsense:
Thank you.

PR #314, opened 2022-03-31
Support optional unique values.

rjmackay:
This was discussed a while ago: https://github.com/sensedeep/dynamodb-onetable/discussions/195 and partially supported. However, while optional values were fine on create, then would fail on update and delete in some cases.

I've tried to cover all the error cases in tests, but it's a little tricky since the tests aren't truly independent.

Builds on https://github.com/sensedeep/dynamodb-onetable/pull/313


mobsense:
Thank you again.

PR #340, opened 2022-05-27
Support types of nested objects

ashsmith:
This change will offer type hints for the object field type where a schema has been defined.

### Existing functionality:

Currently you would see the following when accessing an object type property:
image

And trying to access the street property on an address would cause a type error:
image

### New functionality:

With this change objects that have a schema are now type complete:

image

In a scenario where the `schema` field is not provided the type for an object field it would become `Entity<unknown>` which I think would be the correct behaviour.


mobsense:
Thank you. Appreciated.

phstc:
@ashsmith @mobsense is there an example of how to use `type: object`? 

I could not find any here 👇 

* https://github.com/sensedeep/onedoc/search?q=%22%7B+type%3A+%22
* https://github.com/sensedeep/dynamodb-onetable/search?p=3&q=%22type%3A+%22

Issue #343, opened 2022-06-13
Type safety for array attributes


calebwilson706:
Currently when using typescript array attributes there is not an option to set the type of the individual items of the array which means that when creating a row or getting a row the field is `Any[]` 😞 .

Could there be perhaps an `items:` inside `OneField` so that you can control the item type eg:

```
myField: {
  type: Array,
  items: {
    type: Object,
    schema: {
        //list item schema here
    }
  },
  required: true
},
```

calebwilson706:
I think the problem with a solution like this would be that you would be introducing a breaking change for anyone who would then not define an items type like above

calebwilson706:
```
type EntityFieldFromType<T extends OneField> =
      T['type'] extends (ArrayConstructor | 'array') ? any[]
    : T['type'] extends (BooleanConstructor | 'boolean') ? boolean
    : T['type'] extends (NumberConstructor | 'number') ? number
    : T['type'] extends (ObjectConstructor | 'object') ? Entity<T["schema"]>
    : T['type'] extends (DateConstructor | 'date') ? Date
    : T['type'] extends (ArrayBufferConstructor) ? ArrayBuffer
    : T['type'] extends (StringConstructor | 'string') ? string
    : T['type'] extends (SetConstructor | 'set') ? Set<any>
    : T['type'] extends 'typed-array' ? EntityFieldFromType<T["items"]>[]
    : never;
```    
If you wanted to avoid forcing typing for items on arrays you could add `typed-array` as one of your OneField Types

calebwilson706:
I think I may have worked out a way to do it without an additional OneType, am I free to put up a Pr?

calebwilson706:
Something like this?
```
type EntityFieldFromType<T extends OneField> =
      T['type'] extends (ArrayConstructor | 'array') ? ArrayItemType<T>[]
    : T['type'] extends (BooleanConstructor | 'boolean') ? boolean
    : T['type'] extends (NumberConstructor | 'number') ? number
    : T['type'] extends (ObjectConstructor | 'object') ? Entity<T["schema"]>
    : T['type'] extends (DateConstructor | 'date') ? Date
    : T['type'] extends (ArrayBufferConstructor) ? ArrayBuffer
    : T['type'] extends (StringConstructor | 'string') ? string
    : T['type'] extends (SetConstructor | 'set') ? Set<any>
    : T['type'] extends 'typed-array' ? EntityFieldFromType<T["items"]>[]
[...]

PR #344, opened 2022-06-14
Type safety for arrays

calebwilson706:
Typed items inside arrays are something we really wanted to use in our project, this PR shows how i would implement it, feedback would be appreciated.

### DIFF
--- a/src/Model.d.ts
+++ b/src/Model.d.ts
@@ -17,7 +17,6 @@ export type OneType =
     StringConstructor |
     SetConstructor |
     ArrayBufferConstructor |
-    'typed-array' |
     string;
 
 /*
@@ -104,7 +103,7 @@ type EntityField<T extends OneField> =
     T['enum'] extends readonly EntityFieldFromType<T>[] ? T['enum'][number] : EntityFieldFromType<T>;
 
 type EntityFieldFromType<T extends OneField> =
-      T['type'] extends (ArrayConstructor | 'array') ? any[]
+      T['type'] extends (ArrayConstructor | 'array') ? ArrayItemType<T>[]
     : T['type'] extends (BooleanConstructor | 'boolean') ? boolean
     : T['type'] extends (NumberConstructor | 'number') ? number
     : T['type'] extends (ObjectConstructor | 'object') ? Entity<T["schema"]>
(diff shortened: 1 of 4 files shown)


calebwilson706:
issue here -> 
#343

calebwilson706:
Can update the documentation also, didn't want to spend any time on it until this gets some 👀

mobsense:
Sorry for the delay, but I'm ready to dig into this now.

The changes look pretty good and surgical.

I was wondering would an alternate schema syntax be possible without `items`. i.e. apply the nested `schema` to array elements if the type == Array. E.g.

```
        User: {
            pk:          { type: String, value: '${_type}#${id}' },
            sk:          { type: String, value: '${_type}#' },
            id:          { type: String, generate: 'ulid' },
            accountId:   { type: String },
            name:        { type: String },
            email:       { type: String },

            preferences:    { type: Array, schema: {
                key: { type: String },
                value:    { type: String },
            }}
        }
```

This would be a little simpler and cleaner but perhaps less explicit than your proposed:

```
            preferences:    { type: Array, items: {
                type: Object, schema: {
                    key: { type: String },
                    value:    { type: String },
                }
            }}
```

Please feel free to argue either way.

Michael

calebwilson706:
Yeah I'm happy to go with using the schema attribute, just wondering how this would look like for a list of primitives then? Can we define the schema to be a number just?

mobsense:
Ah, you are right. Your way is necessary.

mobsense:
Thank you. This is a great patch. Do you want to update the docs too?

calebwilson706:
@mobsense yeah i'm happy to update the docs, however i have a pretty full schedule for today and tomorrow so will most likely be the end of the week if thats ok?

mobsense:
Cool. That would be great.

calebwilson706:
@mobsense hello, where all do i have to update the docs, is it just the README.md file or is there other places that need chanegd?

tombailey14:
@mobsense Hi, just wondering if this can be released? Would be very useful to have typed arrays on work I am doing atm. Thanks

mobsense:
@calebwilson706 Docs are here: https://github.com/sensedeep/onedoc

mobsense:
We're just doing some tests and pulling a few issues when using this on some of our live production sites. 

Should be soon though over the next 4 days.

PR #350, opened 2022-07-04
Allow non-required objects with required fields

mikkel-pearl:
At present nested schemas cannot have required fields with a non-required parent. For example

```
{
   parent: {
      type: 'object',
      required: false,
      schema: {
        name:  {
          type: 'string',
          required: true
        }
      }
   }
}
```

Fails validation when parent is undefined.  This minor PR is my best guess at how to resolve this. 

It also has a small fix so it will build on Windows.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "author": "Michael O'Brien <mob@sensedeep.com>",
     "license": "MIT",
     "scripts": {
-        "build": "rm -fr dist/* && tsc -p tsconfig.json && tsc -p tsconfig-cjs.json && ./fixup",
+        "build": "rm -fr dist/* && tsc -p tsconfig.json && tsc -p tsconfig-cjs.json && bash ./fixup",
         "lint": "eslint src",
         "lint-fix": "eslint src --fix",
         "prepare": "npm run build",
--- a/src/Model.js
+++ b/src/Model.js
@@ -1109,7 +1109,10 @@ export class Model {
                     let name = field.name
                     let value = properties[name]
                     if (op == 'put') {
-                        value = value || field.default || {}
+                        value = value || field.default
+                        if (value === undefined && field.required) {
+                            value = {}
+                        }
                     }
                     if (value !== undefined) {
                         rec[name] = rec[name] || value


mobsense:
Hi Mikkel,

Thanks for looking into this.

Just a question, if you have a request nested property, does that not always imply that the parent MUST be required too?

Say we have a parent "info" and a nested property "email".

For email to be present, the item must have "info.email" which requires that "info" exist.

Or am I misunderstanding you?

mikkel-pearl:
In my view 'required' should be dependent on scope, rather than treating it as flat. This would ensure it has alignment with object orientation.

For instance info can be undefined in this Typescript definition. Only if isn't undefined must email exist as a field on it.
```
{
  info?: {
    email: string
  }
}
```
IMO if someone wants to treat attributes as flat but aggregated then they'd use the map feature rather than definining a separate schema under the object type, 

Thoughts?

mobsense:
Consider:
```
info: { type: Object, schema: {
    email: { type: String, required: true},
}}
```

This should implicitly make "info" required and use a default of "{}".
In your patch, you are only setting info to {} if it has a required: true. But if there is not a required: true and email is defined, then it will fail because info will be undefined?

mikkel-pearl:
I'm not clear why you think it should make info required though. Making it required breaks its consistency with object orientation, as demonstrated in the type definition snippet I referred to. The reason I made the PR is exactly because I have a situation where I have an info that isn't required but want to check the schema if info is present.

One of my use cases is having `backgroundImage` where it could be undefined but if it is present then has multiple fields on it, url, height, width, etc. I'm a bit surprised it hasn't arisen beforehand since this is a common use case and one of the stated advantages of using Dynamo!

I mean ultimately it's your call, but treating info as required in your example is both inconsistent with typing but also significantly less expressive. It not only makes it impossible to have optional objects, but also polymorphism. When I have time I'll make a PR that supports polymorphism which is one area that Dynamo has a huge advantage in, but if the spec remains like it is then that won't work.

mobsense:
I think I misspoke and mixed up "required" vs defined as {}. Thanks for persisting in explaining.

mikkel-pearl:
Awesome! Thanks

mobsense:
Thanks for having the passion to contribute and drive this.

shishkin:
I have the same use case and just encountered this fix. Thanks @mikkel-pearl for contributing it!

I think the fix needs to go one step further and ensure that the optional field is not present if empty when reading the item from the DB.

Consider this snippet:

```typescript
  models: {
    User: {
      pk: { type: String, value: "${_type}" },
      sk: { type: String, value: "${name}" },
      name: { type: String, required: true },
      info: {
        type: Object,
        required: false,
        schema: {
          email: { type: String, required: true },
        },
      },
    },
  } as const,
```

```typescript
  await User.create({ name: "Alice" });
  const lot = await User.get({ name: "Alice" });
  expect(lot?.info).toBeUndefined();
```

This test fails as `info` is `{}`. I would argue it needs to adhere to TypeScript typing and ensure it's `undefined` or conforms to the schema.

mikkel-pearl:
Good point @shishkin this is something that I have run into with DynamoDB in the past as well when making an ORM. It really helps to catch these blank objects and make them undefined. If @mobsense isn't on it I might have some time soon

mobsense:
That would be great if you could do that @mikkel-pearl

PR #354, opened 2022-07-08
Add transactions support for removeByFind

rehmanvirani-10p:
Adds transaction support for removeByFind.

### DIFF
--- a/src/Model.js
+++ b/src/Model.js
@@ -626,7 +626,9 @@ export class Model {
             throw new OneTableArgError('Remove cannot retry', {properties})
         }
         params.parse = true
-        let items = await this.find(properties, params)
+        let findParams = Object.assign({}, params)
+        delete findParams.transaction
+        let items = await this.find(properties, findParams)
         if (items.length > 1 && !params.many) {
             throw new OneTableError(`Removing multiple items from "${this.name}". Use many:true to enable.`, {
                 properties,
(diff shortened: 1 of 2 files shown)


src/Model.js
```diff
@@ -626,7 +626,9 @@ export class Model {
             throw new OneTableArgError('Remove cannot retry', {properties})
         }
         params.parse = true
-        let items = await this.find(properties, params)
+        let findParams = Object.assign({}, params);
+        delete findParams.transaction;
+        let items = await this.find(properties, findParams)
```
rehmanvirani-10p, line 631: I was getting `Unknown transaction operation find` error because `removeByFind` is using the same params.

mobsense:
Good catch. Thank you.

rehmanvirani-10p:
@mobsense Thanks for accepting the PR. I see this change in the [v2.4.0](https://github.com/sensedeep/dynamodb-onetable/releases/tag/v2.4.0) tag and it's in pre-release status. Is there something blocking the official release?

mobsense:
Our process is to accumulate changes under the next release tag and let folks test them out for a while before release.  There is nothing blocking release, but nothing forcing either. We try to balance between getting fixes out quickly and having too many releases.

Can you post here if the current repo 2.4 source is working well for you. That feedback helps accelerate the release.

PR #360, opened 2022-07-13
FIX: don't require generated fields on create

mikkel-arturo:
The intellisense is generally great, but isn't quite perfect with a generated field. Right now the only options are
```
{
   id: {
      type: 'string',
      required: true,
      generate: true
   }
}
```
Which will insist on id being present when creating an item. Or 

```
{
   id: {
      type: 'string',
      generate: true
   }
}
```
Which will have a type of string|undefined, making typing ugly throughout the object lifecycle. 

This PR extracts generated fields and makes them optional when creating an item, so that typing works as expected.

### DIFF
--- a/src/Model.d.ts
+++ b/src/Model.d.ts
@@ -130,12 +130,19 @@ export type Optional<T extends OneTypedModel> = {
     -readonly [P in keyof T as T[P]['required'] extends true ? never : P]?: EntityField<T[P]>
 };
 
+export type Generated<T extends OneTypedModel> = {
+    -readonly [P in keyof T as T[P]['generate'] extends true ? P : never]?: EntityField<T[P]>
+};
+
+type ExtractModel<M> = M extends Entity<infer X> ? X : never
+
 /*
     Merge two types
 */
 type Merge<A extends any, B extends any> = {
     [P in keyof (A & B)]: P extends keyof A ? A[P] : (P extends keyof B ? B[P] : never)
 };
+
 
 /*
     Create entity type which includes required and optional types
@@ -261,9 +268,10 @@ export type AnyModel = {
     upsert(properties: OneProperties, params?: OneParams): Promise<AnyEntity | undefined>;
 };
 
+type CreateProperties<T> = Omit<T, keyof Generated<ExtractModel<T>>> | Generated<ExtractModel<T>>
 export class Model<T> {
     constructor(table: any, name: string, options?: ModelConstructorOptions);
-    create(properties: T, params?: OneParams): Promise<T>;
+    create(properties: CreateProperties<T>, params?: OneParams): Promise<T>;
     find(properties?: EntityParametersForFind<T>, params?: OneParams): Promise<Paged<T>>;
     get(properties: EntityParameters<T>, params?: OneParams): Promise<T | undefined>;
     load(properties: EntityParameters<T>, params?: OneParams): Promise<T | undefined>;


dimas-cyriaco:
I was about to create an issue about this. Thank you.

mobsense:
Thank you. Looks good. We'll test it out.

mobsense:
We're having some trouble with this change.

By mistake, I reverted it in the latest release by removing the CreatedProperties<T> in the create() definition in Model.d.ts.

However, when reverting, TS does not catch any missing required fields in create(). 

If I remove CreatedProperties<T>, then missing required fields are flagged, but missing generated fields are also flagged.

Can you check this for me?   I'm using TSC 4.3.2.

mobsense:
The fix may be changing in Model.d.ts:

    type CreateProperties<T> = Omit<T, keyof Generated<ExtractModel<T>>> | Generated<ExtractModel<T>>

to 

    type CreateProperties<T> = Omit<T, Generated<ExtractModel<T>>>

And create() to:

    create(properties: CreateProperties<T>, params?: OneParams): Promise<T>;

mikkel-pearl:
Yes I can confirm. What's weird is that in WebStorm's intellisense it *does* have the correct typing behavior, it's only in the compiler that it isn't caught.

Can you check

```type CreateProperties<T> = Omit<T, keyof Generated<ExtractModel<T>>> & Generated<ExtractModel<T>>``` ?

My original thought was to let the generated keys be optional, which I believe that your proposed change wouldn't allow? Changing the `|` to `&` appears to be working for me

mobsense:
Generated keys can be optional with:

    type CreateProperties<T> = Omit<T, Generated<ExtractModel<T>>>

Not sure why, but it works.

Also works if you put required: true, generated: true. i.e. then it is required to be provided.

mikkel-pearl:
Oh awesome! Such are the mysteries of Typescript typing. Thanks for noticing this

mobsense:
I'd recommend you also check things with VS code for their intellisense.

mikkel-pearl:
Hmm. I was working with a co-worker and tried the ```type CreateProperties<T> = Omit<T, Generated<ExtractModel<T>>>``` line but then it started throwing errors about id being required again.

My original fix ```type CreateProperties<T> = Omit<T, keyof Generated<ExtractModel<T>>> & Generated<ExtractModel<T>>``` was what worked. Can you have another look?

mobsense:
Can you describe your environment. I'm testing with TSC and VS Code.

Your fix did not trigger on missing required properties whereas mine did with VS Code and TSC.

Could this be a WebStorm thing?

mikkel-pearl:
Sorry I don't mean "my original" as in what was in the PR. I meant what I proposed above once you pointed out the issue with the PR.

Here is ```type CreateProperties<T> = Omit<T, keyof Generated<ExtractModel<T>>> & Generated<ExtractModel<T>>``` (notice the change from `|` in the original PR to `&`)

image

You can see it catches the missing communityId in both Webstorm and TSC (v 4.4.3)


Here is ```type CreateProperties<T> = Omit<T, Generated<ExtractModel<T>>>``` as you suggested

image

It is now saying that `id` is required but missing

mobsense:
Is this with VS Code or Web Storm?

mikkel-pearl:
Well the screenshot is webstorm plus tax on my console. My coworker was
using visual studio and got the same results

mobsense:
It is working for me here. So versions please.

PR #369, opened 2022-07-23
DEV: infer getModel type with typesafe input

loetjvr:
I copied the work from PR #308 and applied to latest code to get tests to pass.

I do not recommend merging this yet because it is not backwards compatible. It creates Typescript errors where the model is passed as generic because the infer type is different.

How should we proceed?
Should I try and get previous generic working as well?

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ If you are using the AWS SDK V3, import the AWS V3 `DynamoDBClient` class and th
 Note: you can use the Table.setClient API to defer setting the client or replace the client at any time.
 
 ```javascript
-import Dynamo from 'dynamodb-onetable/Dynamo'
+import {Dynamo} from 'dynamodb-onetable/Dynamo'
 import {Model, Table} from 'dynamodb-onetable'
 import {DynamoDBClient} from '@aws-sdk/client-dynamodb'
 const client = new Dynamo({client: new DynamoDBClient(params)})
--- a/README.md
+++ b/README.md
@@ -246,8 +246,8 @@ let account: AccountType = {
     unknown: 42,           //  Error
 }
 
-//  Get an Account access model
-let Account = table.getModel<AccountType>('Account')
+//  Get an Account access model (<AccountType> is inferred)
+let Account = table.getModel('Account')
 
 let account = await Account.create({
     name: 'Acme',               //  OK
--- a/src/Table.d.ts
+++ b/src/Table.d.ts
@@ -44,7 +44,7 @@ type TableConstructorParams<Schema extends OneSchema> = {
     uuid?: (() => string) | string, //  Function to create a UUID if field schema requires it.
 };
 type ModelNames<Schema> = keyof Schema["models"];
-export class Table<Schema extends OneSchema> {
+export class Table<Schema extends OneSchema = any> {
 
     name: string;
     constructor(params: TableConstructorParams<Schema>);
--- a/src/Table.d.ts
+++ b/src/Table.d.ts
@@ -2,109 +2,194 @@
     Table.d.ts -- Hand crafted type defintions for Table
 */
 
-import { AnyEntity, AnyModel, Model, OneIndexSchema, OneParams, OneProperties, OneModelSchema, OneSchema, Paged, Entity} from "./Model";
+import {
+  AnyEntity,
+  AnyModel,
+  Model,
+  OneIndexSchema,
+  OneParams,
+  OneProperties,
+  OneModelSchema,
+  OneSchema,
+  Paged,
+  Entity,
+} from "./Model";
 
 export type EntityGroup = {
-    [key: string]: AnyEntity[]
+  [key: string]: AnyEntity[];
 };
 
 type TableConstructorParams<Schema extends OneSchema> = {
-    client?: {},                    //  Instance of DocumentClient or Dynamo.
-    crypto?: {},                    //  Crypto configuration.
-    generate?: (() => string),      //  Function to generate IDs for field schema that requires.
-    generic?: boolean,              //  Create a generic (low-level) raw model. Default false.
-    logger?: boolean | ((tag: string, message: string, context: {}) => void),      // Logging callback
-    //  Intercept table reads and writes
-    intercept?: (model: AnyModel, op: string, rec: {}, params: OneParams, raw?: {}) => void,
-    metrics?: boolean | object,     //  Enable CloudWatch metrics.
-    name?: string,                  //  Table name.
-    schema?: Schema,                    //  Table models schema.
-    senselogs?: {},                 //  SenseLogs instance for logging
-    //  Transform record for read / write.
-    transform?: (model: AnyModel, op: string, item: AnyEntity, properties: OneProperties, params?: OneParams, raw?: {}) => AnyEntity,
-    //  Validate properties before writing
-    validate?: (model: AnyModel, properties: OneProperties, params?: OneParams) => {},
-    //  Compute a value for a value template
-    value?: (model: AnyModel, fieldName: string, properties: OneProperties, params?: OneParams) => string,
+  client?: {}; //  Instance of DocumentClient or Dynamo.
+  crypto?: {}; //  Crypto configuration.
+  generate?: () => string; //  Function to generate IDs for field schema that requires.
+  generic?: boolean; //  Create a generic (low-level) raw model. Default false.
+  logger?: boolean | ((tag: string, message: string, context: {}) => void); // Logging callback
+  //  Intercept table reads and writes
+  intercept?: (
+    model: AnyModel,
+    op: string,
+    rec: {},
+    params: OneParams,
+    raw?: {}
+  ) => void;
+  metrics?: boolean | object; //  Enable CloudWatch metrics.
+  name?: string; //  Table name.
+  schema?: Schema; //  Table models schema.
+  senselogs?: {}; //  SenseLogs instance for logging
+  //  Transform record for read / write.
+  transform?: (
+    model: AnyModel,
+    op: string,
+    item: AnyEntity,
+    properties: OneProperties,
+    params?: OneParams,
+    raw?: {}
+  ) => AnyEntity;
+  //  Validate properties before writing
+  validate?: (
+    model: AnyModel,
+    properties: OneProperties,
+    params?: OneParams
+  ) => {};
+  //  Compute a value for a value template
+  value?: (
+    model: AnyModel,
+    fieldName: string,
+    properties: OneProperties,
+    params?: OneParams
+  ) => string;
 
-    // https://www.npmjs.com/package/dataloader DataLoader constructor
-    dataloader?: new (batchLoadFn: any, options?: any) => any
+  // https://www.npmjs.com/package/dataloader DataLoader constructor
+  dataloader?: new (batchLoadFn: any, options?: any) => any;
 
-    //  DEPRECATED 2.3 - Should now be specified via the schema.params
-    createdField?: string,          //  Name of "created" timestamp attribute.
-    hidden?: boolean,               //  Hide key attributes in Javascript properties. Default false.
-    isoDates?: boolean,             //  Set to true to store dates as Javascript ISO Date strings.
-    nulls?: boolean,                //  Store nulls in database attributes. Default false.
-    timestamps?: boolean,           //  Make "created" and "updated" timestamps. Default true.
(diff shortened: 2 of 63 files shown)


loetjvr:
I was able to allow the previous generic type as well. So everything should work as expected now.

loetjvr:
Ok I see here we have test:v2 and test:v3 which I didn't run cause CONTRIBUTING.md only indicated test. I will have a look at these. I was also using docker. Not sure if that changes the results .

loetjvr:
Seems my tests we returning cached results so was not seeing all the type errors. The type errors should pass now but I did see some timeout errors which I am not sure if its just my machine

loetjvr:
a problem I see in the tests are that the model and schema gets modified further down the chain which then doesn't work with Typescript cause it compares with the schema that is sent in the table constructor. Not sure if this is test specific behaviour or something that needs to work that way. I am not sure how Typescript can keep track of updated schema

mobsense:
Can you point me to a case of this so I can better understand.

We could change the tests to not modify the schemas. That would probably be a good direction.

test/table-constructor.ts
```diff
@@ -7,7 +7,8 @@ import {DefaultSchema} from './schemas'
 // jest.setTimeout(7200 * 1000)
 
 let TableName = 'TableConstructorTestTable'
-let table: Table = null
+// Schema need to be passed because its not in the constructor. Using any because it gets modified
+let table: Table<any>
```
loetjvr, line 11: @mobsense here I had to pass any in the schema cause it gets modified

test/schemas/tenantSchema.ts
```diff
@@ -21,7 +21,7 @@ export default {
         User: {
             pk:         { type: String, value: 'Account#${accountId}' },
             sk:         { type: String, value: '${_type}#${id}' },
-            accountId:  { type: String, required: true },
+            accountId:  { type: String },
             id:         { type: String, generate: 'ulid' },
```
mobsense, line 25: Why remove required here?  As it is part of the PK, it should be required.

README.md
```diff
@@ -246,8 +246,8 @@ let account: AccountType = {
     unknown: 42,           //  Error
 }
 
-//  Get an Account access model
-let Account = table.getModel<AccountType>('Account')
+//  Get an Account access model (<AccountType> is inferred)
+let Account = table.getModel('Account')
```
mobsense, line 251: This is much cleaner.

src/Table.d.ts
```diff
@@ -2,13 +2,13 @@
...
-import { AnyEntity, AnyModel, Model, OneIndexSchema, OneParams, OneProperties, OneModelSchema, OneSchema, Paged} from "./Model";
+import { AnyEntity, AnyModel, Model, OneIndexSchema, OneParams, OneProperties, OneModelSchema, OneSchema, Paged, Entity} from "./Model";
 
 export type EntityGroup = {
     [key: string]: AnyEntity[]
 };
 
-type TableConstructorParams = {
+type TableConstructorParams<Schema extends OneSchema> = {
     client?: {},                    //  Instance of DocumentClient or Dynamo.
```
mobsense, line 12: Unfortunate that this is breaking compatibility. So we can't introduce this in a 2.x.

However, our plans are for 3.0 to be based on *.ts files. So this could be a good basis for that work.
We're trying to clear the decks and get a very stable 2.x before we start the work.

Are you interested in using this PR and we could start a 3.x branch?

test/context.ts
```diff
@@ -110,7 +110,7 @@ test('Get single user', async() => {
 
 test('Remove many users (returning ALL_OLD)', async() => {
     //  PK comes from context
-    let removed = await User.remove({}, {many: true})
+    let removed: any = await User.remove({}, {many: true})
     expect(removed).toHaveLength(3)
```
mobsense, line 114: Why is the any required here?

test/inline-model-typescript.ts
```diff
@@ -9,7 +9,8 @@
  */
 import {AWS, Client, Entity, Model, Table, dump, print} from './utils/init'
 
-const table = new Table({
+// send any schema because it gets modified
+const table = new Table<any>({
     name: 'InlineModelTypeScriptTestTable',
```
mobsense, line 14: Yes -- we should cleanup the any as you say.

mobsense:
I've posted some review comments on the changed files.

Given this is breaking compatibility (table constructor), we'll need to defer to our 3.x branch which we plan to explore converting the source to TS. What are your thoughts/

loetjvr:
The changes I introduced based on the previous open PR is just adding types so I think we should be good on the backwards compatibility part. I think it would be a really good gain if we can be sure it works.

The areas where is doesn't work is if the schema is not sent via the table constructor. Would it be an actual use case where the table instance is created without the schema? If so then yes we have a problem. This could maybe be solved with typescript but it might take a lot of work as I am also just learning advanced library typescript.

It might also be possible to switch over to typescript in the repo with exactly the same functionality thus keeping us in a v2. It will be a nicer transition to v3 and there will be gains with type safe implementations in the repo.

I will go through the other questions and reply to those. On the ones where I changed tests that seem not to be related was because I ran into type issues but will confirm.

test/table-constructor.ts
loetjvr: Ok am I now using any in the type if its not provided here. Again this will only happen if the table is created without a schema

test/schemas/tenantSchema.ts
loetjvr: This issue is now visible because the model is now inferred so it means that in contexts.ts we cant create a user without the account id so I just made it not required. The alternative is the fix the create function here https://github.com/sensedeep/dynamodb-onetable/blob/main/test/context.ts#L76

test/context.ts
loetjvr: again here its making an issue visible because the model is inferred. It seems that based on the model type its only expecting the entity returned but here we using an array. So the many is not recognised. We can explore to see if we can get this working in typescript but this is issue was not caused by the new code

test/inline-model-typescript.ts
loetjvr: not sure why its not showing the update here but based on the same change above. The schema now defaults to any if not provided

src/Table.d.ts
mobsense: Will TS users who construct their table using a bare `new Table()` work or do they have to change their code to be `new Table<XX>()`?

test/schemas/tenantSchema.ts
mobsense: Okay, that makes sense if the accountId is coming from the context.

test/context.ts
mobsense: Yes, we should update the remove() signature to allow Array<T>

test/inline-model-typescript.ts
mobsense: Okay, understand the issue now.

mobsense:
The use case of not providing the schema is pretty rare. It was in the API from the start, but the capabilities of the schema have grown and I don't know of anyone using modified schemas. It certainly breaks tools like SenseDeep.

So I don't mind it being ugly, and losing some Typing in the case where the schema is not provided to the Table constructor.

Please see the changes we're discussing in #372 regarding create() typing.

src/Table.d.ts
loetjvr: The Schema type here is inferred so existing code should just work its not really meant to pass down generic types.

test/context.ts
loetjvr: ok so this should be in a new PR right?

mobsense: Yes, that would be best. 

The key open question for me is the backwards compatibility. i.e.

>Given this is breaking compatibility (table constructor), we'll need to defer to our 3.x branch which we plan to explore converting the source to TS. What are your thoughts?

loetjvr: this already works like this though right? the only reason it's visible now is because models are inferred it is always typed. So if someone in the current version created a typed model they would see the same issue no?

mobsense: I may be getting confused.

Can you give me a little sample of exactly how the new TableConstructorParmams works?

Before

    type TableConstructorParams = {

Proposed

    type TableConstructorParams<Schema extends OneSchema> = {

loetjvr: This is also new to me. I followed what was done on the previous PR to get it to work. So the goal was to get the models that are passed in the schema. To get access to this schema we pass an inferred type from table class to the constructor. The inferred schema type can then be used to expose the models to the model method. So the schema here is just a reference that gets populated from the constructor. Its not an expected type that needs to be explicitly passed.

mobsense:
Thank you so much for your efforts on this.

To push forward, can you resolve the conflicts with Table.d.ts so I can merge.

I'll pull the PR and test it out locally and try to resolve the backwards compatibility issue and then see if we can get this merged.

loetjvr:
No problem glad to help as I am starting a few projects using this repo. I fixed the conflict.

mobsense:
The src/Table.d.ts has a lot of formatting changes. I'd like to revert those so we can focus on only the changes. We'll need to stick to the prior formatting before we merge.

Anyway, I'll test it out this weekend and give it a whirl. Looking forward to this.

loetjvr:
bleh ok I will fix these test issues

loetjvr:
> The src/Table.d.ts has a lot of formatting changes. I'd like to revert those so we can focus on only the changes. We'll need to stick to the prior formatting before we merge.
> 
> Anyway, I'll test it out this weekend and give it a whirl. Looking forward to this.

formatting changes like tab space etc? I might have accidentally saved with default prettier. Could we add something https://editorconfig.org/ to set these for editors?

loetjvr:
fixed the formatting. I did discover that the model suggestion is not working. So you could add any model and its not type checking that the model exists. It was probably because of the backwards compatibility I added. I work on this again

loetjvr:
fixed inference again and you will see in the crud test we correctly getting a typescript error when model not known.

mobsense:
Thank you again for your work on this. I understand this can be a bit tedious, but we're getting close.

I've checked out the PR vs the current code:

Observations:

* Builds and tests nicely -- thank you!
* Lints cleanly
* Formatting is good and I could easily compare with the current code (Thank you again).

One issue:

* The current version will detect missing required parameters in the create() API. The PR does not.

I manually applied the change for the create() typing from the current branch which does the required checking. When added to the PR, the missing required parameters are still not checked.

```
type CreateProperties<T> = Omit<T, Generated<ExtractModel<T>>>

export class Model<T> {
    create(properties: CreateProperties<T>, params?: OneParams): Promise<T>;
```

In the test below, I comment out the required attribute: "name".

The screen shot for the current code is:

Screen Shot 2022-08-06 at 11 19 54 AM

The screen shot for the PR is:
Screen Shot 2022-08-06 at 11 25 23 AM

You can see the type for create() in the current code is CreateProperties<UserType> whereas the PR is very different.

The problem may be the getModel definition
```
67c68
<     getModel<T>(name: string): Model<T>;
---
>     getModel<T>(name: ModelNames<Schema>): Model<Entity<Schema["models"][T]>>;
```

The test schema is:

```
const Schema = {
    format: 'onetable:1.1.0',
    version: '0.0.1',
    indexes: {
        primary: { hash: 'pk', sort: 'sk', project: 'all' },
        gs1: { hash: 'gs1pk', sort: 'gs1sk', project: 'all' },
    },
    models: {
        User: {
            pk:          { type: 'string', value: '${_type}#' },
            sk:          { type: 'string', value: '${_type}#${id}' },

            gs1pk:       { type: 'string', value: '${_type}#' },
            gs1sk:       { type: 'string', value: '${_type}#${id}' },

            id:          { type: 'string', generate: 'ulid' },
            name:        { type: 'string', required: true },
            email:       { type: 'string', },
            version:     { type: 'number', required: true },
            address:     { type: Object, schema: {
                street:  { type: 'string', required: true },
                city:    { type: 'string' },
                state:   { type: 'string' },
                zip:     { type: 'number' },
            }},
        }
    } as const,
    params: {
        partial: true,
    }
}
```

The code is:
```
    user = await User.create({
        // name: 'Michael', 
        // id: '22',
        email: 'user@example.com',
        address: {
            street: '42 Park Ave', 
            state: 'WA',
            zip: 98102,
        },
        version: 44
    }, {log: true})
```

loetjvr:
Ok fixed the infer type property check and also added tests so we don't have regression.

mobsense:
Thank you. I've got some manual merges to do here. Will post back status in a few days.

PR #371, opened 2022-07-27
Break table definition out as its own function

volkipp:
Since the Table class already generates a definition description with createTable, it would be helpful to expose that functionality as a standalone method. When using infrastructure as code frameworks such as AWS CDK, Serverless, or Serverless Stack, this would be a great benefit to quickly generate the resource definition from a Onetable schema and table name. 

This pull request breaks the definition generation functionality of `Table.createTable()` out as `Table.getTableDefinition()`

Currently, the only way to do this programmatically is to mock a fake dynamodb client.

**Before:**
```javascript

const table = new Table({
  name: 'DynamoTable',
  schema: mySchema,
  // Stub in a createTable used to describe the table
  client: {
    V3: true,
    createTable(def) {
      return def
    },
  },
})

const definition = await table.createTable()
```

**After:**

```javascript

const table = new Table({
  name: 'DynamoTable',
  schema: mySchema,
})

const definition = table.getTableDefinition()
```

### DIFF
--- a/src/Table.d.ts
+++ b/src/Table.d.ts
@@ -54,6 +54,7 @@ export class Table {
     batchGet(batch: any, params?: OneParams): Promise<{}[]>;
     batchWrite(batch: any, params?: OneParams): Promise<{}>;
     clearContext(): Table;
+    getTableDefinition(params?: {}): {};
     createTable(params?: {}): Promise<{}>;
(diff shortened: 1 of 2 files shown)


mobsense:
Looks good. Thank you.

PR #375, opened 2022-07-28
Fixes scan/find count aggregation when paging

tautologistics:
When passing the option `count: true` to `find()`/`scan()` the count results are not aggregated when the record set exceeds DB limits and requires paging to get all record counts. Instead, the count result of the last page is returned.

This fix simply adds a `count` aggregation variable to `run()` and uses that as the return result instead of `result.Count`, which is only the count of records from the last query.

Note: the test required creating a bunch of large records to exceed DynamoDB's scan size limit to ensure `run()` has to page through results. It's not that bad but it may make sense to flag this test off?

### DIFF
--- a/src/Model.js
+++ b/src/Model.js
@@ -346,7 +346,7 @@ export class Model {
         /*
             Run command. Paginate if required.
          */
-        let pages = 0, items = []
+        let pages = 0, items = [], count = 0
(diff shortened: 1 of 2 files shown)


mobsense:
Thanks. appreciated.

PR #380, opened 2022-08-01
i had this issue when trying it on my machine. this change fixed it

CrackedPixel:
```
const Dynamo = require('../node_modules/dynamodb-onetable/Dynamo');

const test = new Dynamo();
```

the above would give me the following error: `TypeError: Dynamo is not a constructor`. However, it was resolved with this small change
```
const {Dynamo} = require('../node_modules/dynamodb-onetable/Dynamo');

const test = new Dynamo();
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ If you are using the AWS SDK V3, import the AWS V3 `DynamoDBClient` class and th
 Note: you can use the Table.setClient API to defer setting the client or replace the client at any time.
 
 ```javascript
-import Dynamo from 'dynamodb-onetable/Dynamo'
+import {Dynamo} from 'dynamodb-onetable/Dynamo'
 import {Model, Table} from 'dynamodb-onetable'
 import {DynamoDBClient} from '@aws-sdk/client-dynamodb'
 const client = new Dynamo({client: new DynamoDBClient(params)})


mobsense:
But Dynamo.js has an export default at the bottom.

You should be able to do:

import Dynamo from 'dynamodb-onetable/Dynamo'

Why were you using ../node_modules.....?

mobsense:
Please let me know your thoughts. I want to make sure this PR is required by other users and won't break things.

The import as was, works fine for us here. So we need to understand why it doesn't work for you.

CrackedPixel:
> But Dynamo.js has an export default at the bottom.
> 
> You should be able to do:
> 
> import Dynamo from 'dynamodb-onetable/Dynamo'
> 
> Why were you using ../node_modules.....?

Should be able to, yes. I was scratching my head for about an hour wondering why it wouldnt work even though it was default.

Due to the project structure, packages are stored in a separate folder (so that other functions can all use them, and have a single folder location after deployment). Because of this, we access the node_modules folder directly instead of using package names, ex `require('../node_modules/something')` vs `require('something')`

Let me see if i can do a super small example and dig into this more, isolated from my actual project

CrackedPixel:
@mobsense 

Figured it out! Using ES6 imports (`import Dynamo from 'dynamodb-onetable/Dynamo');`) works perfectly fine with `type: module` in package.json, but my project is using ES5 imports (`const Dynamo = require('dynamodb-onetable/Dynamo');`) and it does NOT work, it needs to be { Dynamo }

Sample project:
```
npm i dynamodb-onetable/Dynamo
```
app.js
```
const Dynamo = require('dynamodb-onetable/Dynamo');
const client = new Dynamo();
```
```
node app.js
```

it will crash

changing app.js, it will pass
```
const { Dynamo } = require('dynamodb-onetable/Dynamo');
const client = new Dynamo();
```

mobsense:
Thank you for this.

Issue #393, opened 2022-09-10
TypeScript 2.5.0 error TS2536: Type '"models"' cannot be used to index type 'Schema'.


altrr2:
Unable to build using tsc after upgrading to 2.5.0, the following comes up:

```
npm run build

> @dream-team/one-table@2.25.0 build
> tsc

../../node_modules/dynamodb-onetable/dist/mjs/Table.d.ts:51:33 - error TS2536: Type '"models"' cannot be used to index type 'Schema'.

51 type ModelNames<Schema> = keyof Schema["models"];
                                   ~~~~~~~~~~~~~~~~


Found 1 error in ../../node_modules/dynamodb-onetable/dist/mjs/Table.d.ts:51
```
Downgrading to 2.4.0 resolves the problem.

Using:
Node 16 LTS
"typescript": "^4.8.3"

mobsense:
Please post a reproducible test file (debug.ts) that we can use to replicate.

Depends on your schema.

rcoundon:
This feels similar to what I'm experiencing, logged as #391

mobsense:
See update in #391

skrud-dt:
I think this is different than #391 . My experience is is that this is happening when TypeScript is evaluating the type declaration files for OneTable.

```
 ../../../node_modules/dynamodb-onetable/dist/mjs/Model.d.ts(105,65): error TS2344: Type 'T["schema"]' does not satisfy the constraint 'OneModel'.
   Type 'OneModel | undefined' is not assignable to type 'OneModel'.
     Type 'undefined' is not assignable to type 'OneModel'.
 ../../../node_modules/dynamodb-onetable/dist/mjs/Model.d.ts(110,61): error TS2344: Type 'T["items"]' does not satisfy the constraint 'OneField'.
   Type 'OneField | undefined' is not assignable to type 'OneField'.
     Type 'undefined' is not assignable to type 'OneField'.
 ../../../node_modules/dynamodb-onetable/dist/mjs/Table.d.ts(51,33): error TS2536: Type '"models"' cannot be used to index type 'Schema'.
```

None of the code in my library is actually referenced here. The only way I've found to get around this so far is by adding `skipLibCheck: true` to `tsconfig`.

I think this is happening because `T['schema']` is defined on `OneField` like this:
```
    schema?: OneModel,
```
https://github.com/sensedeep/dynamodb-onetable/blob/main/src/Model.d.ts#L53

[...]

PR #397, opened 2022-09-20
fix(types): fix type error on models and table classes

skrud-dt:
This change fixes the type declarations for the Model and Schema so that they _exclude_ `undefined` values. This is referenced in #393.

The error can be reproduced with a minimal project, here: https://github.com/skrud-dt/dynamodb-onetable-types-reproduce

The reason this is happening is that `'schema'` is defined on the `OneField` model as possibly being `undefined`, which causes TypeScript to complain about it. The workaround here is to just use `Exclude<T["schema"], undefined>` so that we're explicitly saying that "we want the type of `T["schema"]` that's not undefined".

To show that this works I fixed it in my reproduction example in this branch, using a patch. https://github.com/skrud-dt/dynamodb-onetable-types-reproduce/compare/main...fix-with-patch


mobsense:
Thank you. I'll merge and test on all our sites and test-suite.

PR #400, opened 2022-10-10
Support for Converting DynamoDb stream records to New and Old model objects

AlexStansfield:
I needed to be able to take DynamoDB streams and get the OneTable model objects from the records.

The code should support New Image, Old Image and New And Old Image view types. Keys only is not supported but I guess in theory code could be added to fetch the item from dynamodb, I didn't need this so didn't build it.

I found the `Table.unmarshall` method for sdk v2 didn't support the format of the stream record so the extra `unmarshallStreamImage` method I added inside the `stream` method

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,6 +13,7 @@
                 "@aws-sdk/util-dynamodb": "^3.188.0"
             },
             "devDependencies": {
+                "@types/aws-lambda": "^8.10.108",
                 "@types/jest": "^29.1.2",
                 "@types/node": "^18.8.5",
                 "aws-sdk": "^2.1232.0",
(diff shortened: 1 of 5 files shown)


mobsense:
Thank you. Sounds like a good addition. Let us check it out before merging.

AlexStansfield:
Great, let me know if I missed anything or did anything not in the coding style of the project and I'll adjust.

AlexStansfield:
Hi, found an issue when i tried in production that it was getting issues from the "Unique" records that are also stored in the table.

I've already fixed with a check that anything it's converting belongs to one of the table model. I'll add a test for this case and then update the PR.

mobsense:
Sorry -- a release conflicted the package.json.

Can you please resolve and resubmit.

AlexStansfield:
Sorry for the delay, I've fixed the conflict and added the fix for non-model images in the stream

mobsense:
Thank you for this patch. A nice addition.

mobsense:
We'll let users who pull directly from GitHub test this update first and will include in the next NPM release.

AlexStansfield:
Thanks so much!

PR #405, opened 2022-10-20
fix: OneField.encode typescript definition needs to allow for (string|number)

erik-dreamteam:
per the docs, encode is supposed to follow a format of ['sk', '#', 1] which means it's a mix of strings and numbers


mobsense:
Thanks for this. Let me know when you are ready for me to review.

erik-dreamteam:
Please review

mobsense:
Thanks

PR #408, opened 2022-10-24
fix: Allow encode in typescript to use a RegExp for the deliminiter

erik-dreamteam:
For javascript it just implicitly allows that as an option to the split method.


erik-dreamteam:
please review

mobsense:
Done. Thank you again.

PR #409, opened 2022-10-25
feat: add support for ConditionCheck transaction items

erik-dreamteam:
This enables writing conditional expressions on the transaction so that it fails if the conditions aren't met


erik-dreamteam:
relates to https://github.com/sensedeep/dynamodb-onetable/pull/409

mobsense:
Erik, you are on a roll!!!

Awesome.

PR #412, opened 2022-11-04
fix: add missing scope property to OneModel typescript definition

erik-dreamteam:


mobsense:
Thanks -- easy merge that one.

PR #422, opened 2022-11-17
fix: passing `keyof Schema["model"]` to `table.getModel` caused type-error

Haringat:
Fixes issues like "TS2345: Argument of type 'string | number' is not assignable to parameter of type '(string extends keyof TSchema["models"] ? keyof TSchema["models"] & string : keyof TSchema["models"]) | (number extends keyof TSchema["models"] ? keyof TSchema["models"] & number : keyof TSchema["models"])'.   Type 'string' is not assignable to type '(string extends keyof TSchema["models"] ? keyof TSchema["models"] & string : keyof TSchema["models"]) | (number extends keyof TSchema["models"] ? keyof TSchema["models"] & number : keyof TSchema["models"])'." by splitting up the cases and make it easier for typescript to identify whether we want to use the type parameter for only declaring the model we wish to receive or use it to specify the name of the model in the schema which it should fetch from there.

### DIFF
--- a/src/Table.d.ts
+++ b/src/Table.d.ts
@@ -79,7 +79,8 @@ export class Table<Schema extends OneSchema = any> {
     generate(): string;
     getLog(): any;
     getKeys(): Promise<OneIndex>;
-    getModel<T>(name: T extends ModelNames<Schema> ? T : ModelNames<Schema>): T extends string ? Model<Entity<Schema["models"][T]>> : Model<Entity<ExtractModel<T>>>;
+    getModel<T extends ModelNames<Schema>>(name: T): Model<Entity<Schema["models"][T]>>;
+    getModel<T>(name: ModelNames<Schema>): Model<Entity<ExtractModel<T>>>;
     getCurrentSchema(): {};
     groupByType(items: AnyEntity[], params?: OneParams): EntityGroup;
     listModels(): AnyModel[];


mobsense:
Thanks for this. I wan't aware of using multiple type declarations for a single method.

Can you explain a bit more. This is very brittle code with lots of use cases. It is easy (way too easy) to break other use cases.

More detail will help and a test piece of code that fails with your error message.

Haringat:
As far as I understand the previous code, there were two distinct uses for T:
1. (as a string): T must extend keyof Schema["models"] and the method returns the Model of the corresponding part of the schema model.
2. (as an object): T is an arbitrary object (within the constraints of the type parameter of ExtractModel) and the name parameter describes the name of the model inside the schema. Now T is directly passed to ExtractModel and the type is not extracted from the schema.

And those are the two use-cases I split the method into.

mobsense:
Great, thank you.

What was your use case that triggered the error message. Can you give a code sample of the error?

Haringat:
Sadly, I cannot do that.
That was code from my day-job and my employer probably would not like me to leak code from there online.

mobsense:
Understand, but can you describe or use pseudo code to show what caused it.

I'm just trying to reproduce the issue and need to do that before we change this pretty brittle TS code.

Haringat:
It was roughly something like this:
```ts
function x<TSchema extends OneSchema>(schema: TSchema) {
    const table = new Table({
        schema
    });
    return {
        getEntity(entityName: keyof TSchema["models"]): Entity<TSchema["models"][entityName]> {
            return table.getEntity(entityName);
        }
    };
}
```

mobsense:
We'll defer this to the TS conversion work going on.

I've added your change as a comment to the existing code.

PR #448, opened 2023-02-16
fix: replace sdk v2 unmarshaller

AlexStansfield:
From this discussion - https://github.com/sensedeep/dynamodb-onetable/discussions/445#discussioncomment-4987730

I've removed the use of the sdk v2 converter. As the stream records are the same structure no matter what sdk you use I've opted to use the v3 unmarshall method as there is already a hard dependency of the v3 library that contains the unmarshall method and it performs the same as the v2 version.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -17,7 +17,6 @@ module.exports = {
         '!**/helpers.[jt]s',
         '!**/schemas/*',
         '!**/proto.ts',
-        '!**/stream.ts',
     ],
     globalSetup: '<rootDir>/test/utils/setup.ts',
     globalTeardown: '<rootDir>/test/utils/teardown.ts',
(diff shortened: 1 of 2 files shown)


dev-embedthis:
Thanks for the quick turn around.

Question: won't that create a hard dep on @aws-sdk?

I.e. hard dep on the V3 sdk?

AlexStansfield:
> Thanks for the quick turn around.
> 
> Question: won't that create a hard dep on @aws-sdk?
> 
> I.e. hard dep on the V3 sdk?

Based on the package.json dependencies it looked like the hard dep already existed

Screenshot 2566-02-16 at 10 49 16

dev-embedthis:
You are right. Perhaps for the V2 folks, we've been pulling in via Dynamo.js the V3 SDK, but tree shaking eliminates it if not used.

So that looks fine. I'll merge.

Given AWS are not aggressively deprecating the V2 SDK, we'll make the library future versions at some point, V3 only.

dev-embedthis:
Thank you!

PR #451, opened 2023-02-26
Add 'start' as valid key in OneParams options

tqhoughton:
Spent a few hours trying to figure out what the option is that corresponds to the `ExclusiveStartKey` in regular dynamo, after searching for a bit I found [this blog post](https://www.sensedeep.com/blog/posts/2021/dynamodb-onetable-tour.html) that made me discover the `start` parameter.

Figured this would help out people who rely on the type definitions to know what is possible like I tend to do

### DIFF
--- a/src/Model.d.ts
+++ b/src/Model.d.ts
@@ -268,6 +268,7 @@ export type OneParams = {
     segments?: number
     select?: string
     set?: object
+    start?: object
     stats?: object
     substitutions?: object
     timestamps?: boolean


tqhoughton:
After further testing I don't think this actually defines the ExclusiveStartKey, my bad. I've tried `next` and `prev` but when I do so I get errors from the dynamodb marshaller that undefined values are being passed in to a map, set or list.

dev-embedthis:
Sorry, that blog post should be updated.

The params are next/prev.

If you are getting errors, please post them to discussions or issues and we'll investigate. A test case really helps.

PR #461, opened 2023-03-16
Allow encoding of attributes to automatically fetch values without knowing separator or position

AlexStansfield:
This is still work in progress but just thought I'd put up what I got so far to see if there is any interest.

I was looking at using the encode feature to reduce attributes and have their values fetched from value template attributes.

The issue I had was I have two delimiters `#` and `:`, I use the `#` to denote the next following is an ID and then I use `:` for everything else. 

Example:
```
Account#${accountId}:Company#${companyId}:Product#${productId}:Batch#${batchId}
```

With the existing functionality this would not work with encode.

I wondered if it was possible to automatically get attribute values based on the template and the field name.

So this is what I have so far. The way it would work is it would switch to "auto" mode if you don't define a delimiter and position in the encode param but just give the attribute that has the value template.

Of course this would need careful use, for example if your template was `${id}-${anotherId}-${andAnotherId}` and all 3 IDs were UUIDs then it would probably return the wrong values as the `-` everywhere would cause problems with identifying where one placeholder ends and one starts. So it would be up to the developer to make sure they understood the limitations of this and planned accordingly.

I still need to:
- test thoroughly with some real data in a table
- improve the regex in the `getValueFromTemplate` method
- add more unit tests for the `getValueFromTemplate` method

Would appreciate any early feedback on:
- are you even interested in this 
- is the `utils.js` file the right place to put the method (I first put it on Model class but was difficult to unit test)
- my regex-fu is pretty bad, any advice to improving the regex in the method (think it's too restrictive right now)

### DIFF
--- a/src/Model.js
+++ b/src/Model.js
@@ -5,6 +5,7 @@
 */
 import {Expression} from './Expression.js'
 import {OneTableError, OneTableArgError} from './Error.js'
+import {getValueFromTemplate} from './utils.js'
 
 /*
     Ready / write tags for interceptions
@@ -1083,7 +1084,17 @@ export class Model {
             if (value === undefined) {
                 if (field.encode) {
                     let [att, sep, index] = field.encode
-                    value = (raw[att] || '').split(sep)[index]
+                    
+                    if (att && !sep && !index) {
+                        let attField = fields[att]
+                        if (attField && attField.value && raw[att]) {
+                            value = getValueFromTemplate(attField.value, raw[att], name)
+                        } else {
+                            value = undefined
+                        }
+                    } else {
+                        value = (raw[att] || '').split(sep)[index]
+                    }
                 }
                 if (value === undefined) {
                     continue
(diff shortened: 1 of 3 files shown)


AlexStansfield:
Already found that this fails completely when the separators between placeholders is always the same (e.g `${accountId}#${companyId}#${productId}#${batchId}`). Looking to see if there is anyway I can support that but this seems like exactly the sort of case the existing `encode` functionality would be perfect for.

AlexStansfield:
So I realised that the way I was doing it would probably just cause headaches. I've added a new method `getValuesFromTemplate` with tests gets all the values from a template and it's value.

This seems more robust, the only issue is if you're using simple delimiters and the values in the placeholders will also contain it. (e.g using `-` and having UUIDs).

The main issue here is that the way I'm currently using it then you'd be calling it for every value you want to pull out of the template. I'd look at storing it once parsed the first time so we can easily get values again should we have another attribute that has it's value encoded in the same template.

Still work in progress but think I'll stop here and wait for feedback just in case this isn't something you'd want.

AlexStansfield:
closing for now, haven't had the time to really work this into a proper solution, might come back to it later

mobsense:
Okay, understand.

PR #469, opened 2023-04-17
fix: add back import of `unmarshall` as required in `unmarshallStreamImage`

AlexStansfield:
Seems in the most recent commit on `main` this import was removed which breaks the build as it's required for support for stream images in sdk v2.

### DIFF
--- a/src/Table.js
+++ b/src/Table.js
@@ -12,6 +12,7 @@ import {Expression} from './Expression.js'
 import {Schema} from './Schema.js'
 import {Metrics} from './Metrics.js'
 import {OneTableArgError, OneTableError} from './Error.js'
+import {unmarshall} from '@aws-sdk/util-dynamodb'
 
 /*
     AWS V2 DocumentClient methods


mobsense:
Hi Alex, thanks for this.

One issue: this introduces a hard dep on the V3 SDK. For V2 SDK users, this bloats their deploys.

Also, I see for streams, there is a hard dep on aws-lambda which is not a small package. I'd like to remove these hard deps. 

A better way would be to export it via the Dynamo.js file so it could be referenced via client.unmarshall. Then it would not be a hard dep for V2 users.

I've opened an issue regarding this: #472

AlexStansfield:
Makes sense, I'll update

AlexStansfield:
Issue seems to have been fixed already so closing this

mobsense:
Alex, sorry to bug you on this.

The function:

```
    unmarshallStreamImage(image, params) {
-        if (!this.V3) {
-            return unmarshall(image)
-        }
        let client = params.client ? params.client : this.client
        let options = client.params.unmarshall
        return client.unmarshall(image, options)
    }
```

has a hard reference to unmarshall from the V3 SDK, but it is used in V2.

For V3, client.params.unmarshall is defined. 

Questions:
- Why is V2 using the V3 unmarshall?
- Is there an alternative?

I want to remove the hard dep on the V3 SDK for those using the V2 SDK.

PR #485, opened 2023-07-10
fix: count: true param issue fix for find method

yousafhanif6186:


mobsense:
The next property must not be enumerable in items[]. It should be accessible, but not be enumerable.

We've applied the change to define count, but can't merge this pull request.

PR #489, opened 2023-07-18
fix: Metrics not including Namespace

rjmackay:
The `metrics` config (line 189) is being shadowed by metrics var (line 208) and so namespace is never included in output


mobsense:
thank you.

BTW: we've got a really cool metrics package about to drop. CloudWatch custom metrics can get very expensive. $5 per active metric. We've got a package basically eliminates this cost. LMK if interested.

PR #494, opened 2023-08-14
WIP/Experimental: Typescript conversion

rjmackay:


src/Dynamo.ts
```diff
@@ -25,8 +26,18 @@ import {
 
 import {marshall, unmarshall} from '@aws-sdk/util-dynamodb'
 
+export interface DynamoParams {
+    client: DynamoDBClient,
+}
+
 export class Dynamo {
-    constructor(params = {}) {
+    client: DynamoDBClient // TODO Confirm type
```
rjmackay, line 34: Is this the right type? I wasn't entirely sure what this wrapper converts from/to

mobsense: That looks right:

import {DynamoDBClient} from '@aws-sdk/client-dynamodb'
const client = new DynamoDBClient(params)

src/Metrics.ts
```diff
@@ -87,23 +109,23 @@ export class Metrics {
         metrics.counters = {}
 
         //  Preserve any prior defined properties functions
-        metrics.properties = metrics.properties || prior.properties
+        metrics.properties = metrics.properties || prior.metrics.properties
```
rjmackay, line 112: Not 100% if this change is right. But looking at the call in Table.js, I think prior is a Metrics instance, so I think this was just buggy before.

src/Metrics.ts
```diff
@@ -54,7 +72,28 @@ const ReadWrite = {
...
+            Table: string,
+            Tenant: string,
+            Source: string,
+            Index: string,
+            Model: string,
+            Operation: string,
+            Profile: string
+        },
+        properties?: Record<string, string>,
+    }
```
rjmackay, line 93: I inferred this from reading the code but not 100% I got this right.

src/UID.ts
```diff
@@ -12,9 +12,9 @@ import Crypto from 'crypto'
 const Letters = '0123456789ABCDEFGHJKMNPQRSTVWXYZZ'
 const LettersLen = Letters.length - 1
 
-export default function UID(size) {
-    let bytes = []
-    let buffer = Crypto.randomBytes(size)
+export default function UID(size: number): string {
+    const bytes = []
+    const buffer = Crypto.randomBytes(size)
```
rjmackay, line 17: eslint just converts these for you helpfully

src/Metrics.ts
mobsense: I think you are right.

src/Metrics.ts
mobsense: Probably should have a defined type rather than inline that. addMetric has a similar inline declaration.

Looks reasonable for now.

mobsense:
Looks like the low hanging fruit is going well.

Be interesting to see how Table, Model and Expression go.

src/Metrics.ts
rjmackay: Yea. I think there's going to be a long tail of clean up. First thing is just to get TS building happily and tests passing. Then deal with that.

rjmackay:
> Looks like the low hanging fruit is going well.
> 
> Be interesting to see how Table, Model and Expression go.

I started on them a little. They'll take a bit longer, and will probably need to be done all in one go because they're so interdependent.

Issue #497, opened 2023-08-18
params.fields are not respected when using table.getItem() or table.queryItems()


ThijsJung:
**Describe the bug**

When querying DynamoDB using `Table.getItem()` or `Table.queryItems()`, `params.fields` are ignored and full objects are returned. When querying using `Model.get()` or `Model.find()`, everything works as expected and only the requested fields are returned.

**To Reproduce**

Full details can be found the [`test/debug.ts` file here](https://gist.github.com/ThijsJung/0a4e5ca4f4a35ea7b1f6be0191541d23). I'll copy/paste an outline of the issue below.

**Cut/Paste**

Schema
```
const schema = {
    version: '0.0.1',
    indexes: {
        primary: {hash: 'pk'},
    },
    params: {
        createdField: 'createdAt',
        updatedField: 'updatedAt',
        isoDates: true,
        timestamps: true,
    },
    models: {
        User: {
            pk: { type: String, value: '${id}'},

            name: { type: String },
            nickName: { type: String },
            id: { type: String, generate: "ulid" },
            email: { type: String, required: true },
        }
    } as const,
}
```

Test
```
const User = table.getModel('User')
const userProperties = {
    name: "Daenerys I Targaryen",
    nickName: "Breaker of chains",
    email: "iHeartDragons@example.com",
}
const user = await User.create(userProperties)
const userId = user.id!

test('model.get() respects the fields.', async () => {
    // Works as expected, only returns id and name.
    const user = await User.get({pk: userId}, {fields: ["id", "name"], parse: true})
    expect(user).toMatchObject({
        name: "Daenerys I Targaryen",
    });
    expect(user!.id).toMatch(Match.ulid);
    expect(user!.nickName).toBeUndefined();
    expect(user!.email).toBeUndefined();
})

test('table.getItem() does not respect the fields.', async () => {
    // Should break but doesn't because the entire record is returned.
    let tableGetItem = await table.getItem({pk: userId}, {fields: ["id", "name"], parse: true})
    expect(tableGetItem).toMatchObject(userProperties);
})
```

**Expected behavior**
`Table.getItem()` and `Table.queryItems()` should respect the `params.fields` and return only the requested fields.

**Environment (please complete the following information):**
 - OS: macOS Venture 13.4.1 (22F82)
 - Node Version: v18.16.0
 - OneTable Version: 2.6.4
 - TypeScript Version: 5.1.6

PR #498, opened 2023-08-18
params.fields are not respected when using table.getItem() or table.queryItems()

ThijsJung:
More details can be found in the issue I filed (#497). I gave it a go and tried to scratch my own itch, let me know what you think. Please feel free to tell me if I forgot anything or if you think there's a better way to fix this issue. Thanks!

### DIFF
--- a/src/Expression.js
+++ b/src/Expression.js
@@ -103,10 +103,17 @@ export class Expression {
                 if (op == 'batchGet') {
                     //  BatchGet params.project must provide attributes not properties
                     this.project.push(`#_${this.addName(name)}`)
-                } else if (fields[name]) {
-                    let att = fields[name].attribute[0]
-                    this.project.push(`#_${this.addName(att)}`)
-                }
+                } else {
+                    // Generic models don't know which attributes exist, so we allow requesting all.
+                    if (this.model.generic){
+                        this.project.push(`#_${this.addName(name)}`)
+                    }else{
+                        if (fields[name]) {
+                            let att = fields[name].attribute[0]
+                            this.project.push(`#_${this.addName(att)}`)
+                        }
+                    }
+                } 
             }
         }
     }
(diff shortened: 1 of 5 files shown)


mobsense:
Thank you for contributing this. Looks pretty good. I'll give it a deeper look over the next few days.

One thing: keep debug.ts unchanged as it is the file for users to modify and submit issues.  Create a new TS file if you need.

Thank you!

ThijsJung:
Thanks for your fast reply! Sure, take your time, and let me know if you see anything that you'd like to see changed. Now that I think about it, I'm not 100% certain to updated all the test that are applicable, so if you know of another method that accepts `fields` that I missed, let me know.

I'll take a moment to revert `debug.ts` and will put mine in a gist instead.

Thanks!

ThijsJung:
By the way, while working on this I ran into a couple of things that are not broken but that might clean things up a bit (1) or make development (IMO) a bit more convenient (2). Let me know if you're interested in a P.R. for these (kinds of) items, I'd be happy to contribute.

Cheers!

1. When [instantiating a genericModel](https://github.com/sensedeep/dynamodb-onetable/blob/3fb91b089efaf4b07aeb1e30c5bb338ad5d30314/src/Schema.js#L151) we pass it both the name (`_Generic`) and an option called (`generic: true`). This can be simplified, I don't think we need the option.
2. While developing I didn't manage to run a test separately (in for instance, in [the `crud` tests](https://github.com/sensedeep/dynamodb-onetable/blob/3fb91b089efaf4b07aeb1e30c5bb338ad5d30314/test/crud.ts). This seems to be because the `user` is created in a separate test which means that that test has to always be run. I also found that if do limit the amount of fields returned, some of my later tests would break because they expected the fields to be present on the user.

ThijsJung:
Perfect, thanks for the update. I'll keep an eye out for the new release. I was happy to help out, keep up the great work!

PR #500, opened 2023-09-15
DOC: change partial default to true

segalaj:
Update `partial` parameter documentation.


mobsense:
Thank you. Good pick up.

Issue #509, opened 2023-11-16
Updating unique template field does not remove unique record


AlexStansfield:
**Describe the bug**

Our user records have email as a unique field. However we do not hard delete user records, instead we soft delete them. We do this by setting the `deletedAt` field to a date.

So for this reason rather than setting the `email` field as unique we have an extra `uniqueEmail` field that uses a template. If there is no `deletedAt` then we want this field to have the email, and if there is a `deletedAt` we want this field to be removed. This should trigger the removal of the unique record for this field.

As this can't be done with a string template we used a function instead.

We're pretty sure this was working before but yesterday I found it no longer did when I went to delete a user and recreate it with the same email.

I did some investigation and found that in `Model.update` when it works out the [`hasUniqueProperties`](https://github.com/sensedeep/dynamodb-onetable/blob/028a644711ad7bd42671531c965a2ff8b2861e37/src/Model.js#L800) the value template function has not been run, so it doesn't know that `uniqueEmail` is now being set to null and so `updateUnique` should be called instead of `updateItem`.

**To Reproduce**

I've created a branch on my fork of this repository - https://github.com/AlexStansfield/dynamodb-onetable/tree/issue/unique-update-with-value-function

Here I added:
* two new fields to the schema, `deletedAt` and `uniqueEmail` - [here](https://github.com/AlexStansfield/dynamodb-onetable/blob/a1a541478b5255864a9b0d8c3a6e9da70c6e66e4/test/schemas/uniqueSchema.ts#L4)
* a function to create the value for `uniqueEmail` (same as `email` if no `deletedAt`, `null` if there is `deletedAt`) - [here](https://github.com/AlexStansfield/dynamodb-onetable/blob/a1a541478b5255864a9b0d8c3a6e9da70c6e66e4/test/unique.ts#L10)
* an extra test to `unique.ts` with the title "Soft delete user and create with same email". - [here](https://github.com/AlexStansfield/dynamodb-onetable/blob/a1a541478b5255864a9b0d8c3a6e9da70c6e66e4/test/unique.ts#L203)

You will see that it fails. It should be expected that the unique record should have been removed so first it fails where it expects the number of items returned in the scan is one less (i.e the unique record is gone). If you remove this expectation you will see the create then fails because it finds the unique record.

The User record however is updated and the `uniqueEmail` field has been removed.

Here is a console.log of the items that it gets from the scan after the "soft delete"

```javascript
[
  {
    sk: { S: '_unique#' },
[...]

PR #510, opened 2023-11-17
Fix issue with Unique Records not being removed when the unique fields are value fields

AlexStansfield:
See Issue #509 for all the details.

I've since taken the tests created for that issue and improved them slightly.

Changes to Model.js:
* Prepare the properties earlier so that we can tell the unique fields have actually changed
* Made a change to the how it determines `hasUniqueProperties` by also checking `params.remove`
* Put a check in `updateItem` so that it only prepares the properties if they haven't already been prepared.
* create new copies of properties and params to pass to `prepareProperties` to avoid issues due to the method mutating the objects causing conflicts if run multiple times


AlexStansfield:
I'm not really happy with the way I had to do this, but because `prepareProperties` mutates the `properties` and `params` in certain situations this caused issues (like adding `remove` twice on fields). 

Personally I think the `prepareProperties` shouldn't mutate these objects and should return new objects with the changes, but that seems like a huge task so this seemed like the way around it. 

If you know a better way I'd be happy to hear them and update accordingly

mobsense:
Alex, thanks for the work on this.

I agree with the comment about mutating properties. 

How is this working for you in practice?

I'll wait for 2 other PRs to have some test, then I'll merge this one.

AlexStansfield:
Actually I was a little worried about the code just because that `prepareProperties` is so complicated that I really couldn't be sure that even though the tests passed this wasn't still messing something up so for now we've taken uniqueness off the schema and doing it only with the existing validation code that checked for duplicates. Not ideal but better than broken data :D

mobsense:
We use uniqueness rarely. It does add a lot of complexity, so we use it for things that really require it like email addresses and account names.

AlexStansfield:
> We use uniqueness rarely. It does add a lot of complexity, so we use it for things that really require it like email addresses and account names.

Our use case is ensuring unique email on user records. The issue for us is we don't want to hard delete users, we want to soft delete them. So a simple unique on `email` field doesn't suffice, no user would be able to sign back up with the same email if their account is deleted.

Solution we found was to use a value function to set a `uniqueEmail` field based on the value of `deletedAt` and have `uniqueEmail` be the unique field. But then we ran into this problem.

If you think my PR is a good fix then I'm happy for you to merge it so we can turn the uniqueness back on.

mobsense:
Alex, sorry for the delay in looking into this. Been swamped with other stuff. 

I'm concerned about the risk of change in the PR -- so that is why I've not just merged it already. I need to dig into it closely.

I'll try to get to this soon.

Michael

AlexStansfield:
> Alex, sorry for the delay in looking into this. Been swamped with other stuff.
> 
> I'm concerned about the risk of change in the PR -- so that is why I've not just merged it already. I need to dig into it closely.

No problem, take your time

mobsense:
Just an update, digging into this now and should have feedback very soon.

mobsense:
Sorry for the delay.

Issue #523, opened 2024-03-01
Cannot Transparently Pass On JSON String From Previously Unncrypted Field


vectorhacker:
**Describe the bug**

When adding encryption to fields that were previously unencrypted the algorithm will fail in the case of finding a stringified JSON object because of the ':' character.

**To Reproduce**

First create a schema with a previously unencrypted field which stores a stringified JSON object. Store some unencrypted JSON strings. Later enable encryption for that field and try to decrypt the records with unencrypted JSON. Usually text should just pass through silently, but this time it will find the ':' characters in the JSON string and throw an error when trying to decrypt.

mobsense:
Hi,

If you change from encrypted to unencrypted or vice-versa, you should perform a database migration to change those fields.

Alternatively, read without a schema and update the time. 

Your PR suggests handling JSON encodings, but there could be any other formats that your PR won't address.

So I'd rather not apply the PR to handle just one such case out of a potential large set of encodings.

PR #524, opened 2024-03-01
FIX: #523 correctly ignore JSON strings in decryption

vectorhacker:
This adds a check for JSON stringified objects before trying to decrypt. Addresses issue #523

### DIFF
--- a/src/Table.js
+++ b/src/Table.js
@@ -71,6 +71,18 @@ process.on(
 )
 
 /*
+    Check if string is valid JSON.
+*/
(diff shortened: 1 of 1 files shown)


mobsense:
Hi,

If you change from encrypted to unencrypted or vice-versa, you should perform a database migration to change those fields.

Alternatively, read without a schema and update the time. 

Your PR suggests handling JSON encodings, but there could be any other formats that your PR won't address.

So I'd rather not apply the PR to handle just one such case out of a potential large set of encodings.

PR #533, opened 2024-07-03
fix: log error on missing required fields when finding with fields selection

### LINKED ISSUE #532: Logging errors emitted for find requests with fields selection not including required fields
I've seen many changes being pushed directly to main and it makes it harder to track the changes.

Right now I have an issue where the lib is logging errors for required fields on count queries which is non-sense and I cannot figure out from where that change came

the first commit I tracked is this one https://github.com/sensedeep/dynamodb-onetable/commit/e20250a6eaf27e4441e25dfdc9797304ed728efe but the message was already there from another commit.

I need help figuring this out.

Also, CI is failing for all pushes directly to main what makes my confidence on the lib quality decrease.

onhate:
fix #532

### DIFF
--- a/src/Model.js
+++ b/src/Model.js
@@ -1136,11 +1136,12 @@ export class Model {
                 }
             } else if (value === undefined) {
                 if (field.required) {
-                    /*  
-                        Transactions transform the properties to return something, but 
-                        does not have all the properties and required fields may be missing)
+                    /*
+                        Transactions transform the properties to return something, but
+                        does not have all the properties and required fields may be missing).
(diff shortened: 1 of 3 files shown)


mobsense:
Thank you!

revmischa:
What should I do if I add a new `required` field but still have old objects in my DB which lack a value for that field? Now this is giving me errors for those objects.

mobsense:
The code will emit a log message but will otherwise work.

You can set the {noerror: true} param to suppress the message in 2.7.5.

The real solution is to do a migration to update old items that now don't match the schema.

PR #535, opened 2024-07-15
Add (multidimensional) enum array typing

thoom76:
### Context
Assume we have the following schema:
```ts
export const TheSchema {
	...,
	models: {
	    SomeModel: {
	        brands: {
	            type: Array,
                    required: true,
	            items: {
                        type: String,
	                enum: Object.values(Brand)
	            }
	        }
	    }
	}
}
```
the following Entity type:
```ts
export type SomeModelEntity = Entity<typeof TheSchema.models.SomeModel>;
```
and the following code:
```ts
const {brands}: SomeModelEntity = {
	...
}
```

### Previously
when we check the type of `brands` we're seeing that it has the type `string[]` instead of the type `Brand[]`

### New
When we check the type of `brands` we'll see that it has type `Brand[]`

### DIFF
--- a/src/Model.d.ts
+++ b/src/Model.d.ts
@@ -100,7 +100,9 @@ export type OneSchemaParams = {
 /*
     Entity field signature generated from the schema
  */
-type EntityField<T extends OneField> = T['enum'] extends readonly EntityFieldFromType<T>[]
+type EntityField<T extends OneField> = T['items'] extends object
+  ? EntityField<T['items']>[]
+  : T['enum'] extends readonly EntityFieldFromType<T>[]
     ? T['enum'][number]
     : EntityFieldFromType<T>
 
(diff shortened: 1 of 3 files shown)


mobsense:
thank you. We'll test this out.

Michael

thoom76:
@mobsense Fixed the typing issue, do you want me to create tests for the typing itself?

mobsense:
Is that possible?

thoom76:
Added the tests :+1:

thoom76:
@mobsense Idk if you guys are aware of it. When simply cloning the repo, making some changes and running `npm run test` locally results in succeeding tests while the CI/CD pipeline fails.

mobsense:
Thank you for the patch.

mobsense:
> @mobsense Idk if you guys are aware of it. When simply cloning the repo, making some changes and running `npm run test` locally results in succeeding tests while the CI/CD pipeline fails on the same changes. Running `nvm use 20 && npm ci && npm run build --if-present && npm run lint --if-present && npm run test:v2 && npm run test:v3` like the pipeline job did the work for me to get the same results.

Not really sure what you mean. The CI/CD seems to be working fine. When does it not work?

PR #542, opened 2024-12-20
Add ability to use a field-specific custom generate function 

bozzaj:
While there is the ability to have a custom generate function at the Table level, there is no way to have something unique to a field. Originally, I was thinking to add support for uuidv7 to the set of built-in ID generators, but this patch just extends the model configuration to support a generate function specific to any field.

Tests are written with a uuidv7-like generate function, but wouldn't necessarily be a replacement for an official uuidv7 library.


dev-embedthis:
Thank you.

bozzaj:
@dev-embedthis I noticed that a 2.7.6 version was generated for this, but that version isn't available as an npm package yet. Does this project have a different process on when the npm packages are published?

See here: https://www.npmjs.com/package/dynamodb-onetable

dev-embedthis:
We are doing final test on this cut before releasing as NPM package. 

We release here first to get some final feedback before releasing on NPM which is our primary delivery vehicle.

dev-embedthis:
Expect release this coming week.

PR #549, opened 2025-02-21
Fix: cannot remove objects by setting to null

AlexStansfield:
I'm not sure how this hasn't affected us before, but today we found that we were unable to remove an object with schema field from an item by setting it to null (as works with scalar fields). Instead it throws an error.

I found two places that caused an issue:
1. `collectNested`: tries to collect the properties of the object but throws an error as it's null so has no properties
2. `convertNulls`: skips the field if it has a schema, so never removes it

I've made this PR to resolve the issue. On my testing it seems to work fine, the unit tests also all pass.

However I'm not sure if this change might have some hidden impact I'm not aware of.


dev-embedthis:
Hi Alex, I'm looking for a new maintainer for this repo.

I've been pulled in a different direction and I was wondering if you'd be interested?

michael

AlexStansfield:
Hi Michael,

Thanks for the offer. I'll need a little time to think about it as I wouldn't want to take something on if I couldn't give it the appropriate focus.

Issue #555, opened 2025-06-24
Typescript does not respect nulls === true on OneField


garymuff:
I noticed typescript does not reflect the correct typing for a potentially null field when `nulls === true`. I made the following change in `src/Model.d.ts` and was able to fix it when nulls is set on a OneField. 

```
type OptionalOrUndefined<T extends OneModel> = {
    -readonly [P in keyof T as T[P]['required'] extends true ? never : P]?: T[P]['nulls'] extends true
        ? EntityField<T[P]> | null | undefined 
        : EntityField<T[P]> | undefined
}
```

There is the case when nulls is set globally on the schema, but there would need additional work to infer the OneSchemaParams which aren't readily available

dev-embedthis:
Thank you

PR #556, opened 2025-06-24
fix: include null typing for nulls field param

garymuff:
This PR addresses https://github.com/sensedeep/dynamodb-onetable/issues/555

Issue: fields do not include a null type when nulls === true. 
Solution: Adjusted `src/Model.d.ts` to checking nulls param, only when required is not true.


dev-embedthis:
Thank you.

